PR #30, opened 2022-12-26
Add lookup entry for AES blackbox function

TomAFrench:
I noticed that we don't support looking up the AES blackbox function so I've added an entry for it and a test to check that we maintain the invariant `BlackBoxFunc::lookup(bb_func.name()) == Some(bb_func)` to help avoid this happening again.

We can automatically track this for new variants by adding the [strum](https://crates.io/crates/strum) crate as a dev-dependency but if we don't want to add this then we can just maintain a static array of all the enum variants.

### DIFF
--- a/acir/src/circuit/blackbox_functions.rs
+++ b/acir/src/circuit/blackbox_functions.rs
@@ -59,6 +59,7 @@ impl BlackBoxFunc {
     }
     pub fn lookup(op_name: &str) -> Option<BlackBoxFunc> {
         match op_name {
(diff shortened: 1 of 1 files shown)


kevaundray:
LGTM, using [EnumIter](https://docs.rs/strum_macros/0.24/strum_macros/derive.EnumIter.html) from strum sounds good since it is a dev-dependency. I'll merge this and create a tracking issue!

kevaundray (approved):

PR #39, opened 2022-12-26
Split off errors from `OpcodeResolution`

TomAFrench:
This implements the changes described in https://github.com/noir-lang/noir/pull/520#discussion_r1052624508.

This is sensible imo as 
- `OpcodeResolution` doesn't make sense for a circuit-level function such as `PartialWitnessGenerator.solve` (what does "Skip" mean when solving a circuit?)
- We can define error messages here in ACVM for the failure modes of solving the circuit. Callers then don't need to understand the reasoning behind all the `OpcodeResolution` variants to be able to separate success vs failure. 

I've lumped in a change to simplify `BlackBoxFuncCall` in the last commit which improves readability imo, happy to revert this if this is unwanted.

### DIFF
--- a/acvm/Cargo.toml
+++ b/acvm/Cargo.toml
@@ -26,6 +26,7 @@ k256 = { version = "0.7.2", features = [
     "arithmetic",
 ] }
 indexmap = "1.7.0"
+thiserror = "1.0.21"
 
 [features]
 bn254 = ["acir_field/bn254"]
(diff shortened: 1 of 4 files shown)


kevaundray:
haha was thinking of doing something similar! This looks good to me overall, I've opened an issue re UnknownError, but its not a blocker for this, so happy to merge

kevaundray (approved):

PR #61, opened 2023-01-20
feat: Add Log Directive

vezenovm:
# Related issue(s)

Resolves log statement issue in noir-lang repo: https://github.com/noir-lang/noir/issues/383

# Description

This includes a Log directive and the logic in the PWG for solving it.

## Summary of changes

I added a Log directive that takes a LogInfo enum. This enum can either be finalized string output or a vector of witnesses whose data has to be fetched when solving the circuit. I then added the logic for printing single values and arrays in the PWG. This logic follows similar logic from acir gen when we can construct a finalized output string.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/directives.rs
+++ b/acir/src/circuit/directives.rs
@@ -2,7 +2,7 @@ use std::io::{Read, Write};
 
 use crate::{
     native_types::{Expression, Witness},
-    serialisation::{read_n, read_u16, read_u32, write_bytes, write_n, write_u16, write_u32},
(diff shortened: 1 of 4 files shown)


kevaundray:
Out of scope, but noting for reference: We won't be able to have the equivalent of templating just yet. ie `format!(" x: {value}")`

For that LogInfo would probably need to be Recurisvely defined. We can open up a future issue/PR for this

kevaundray (approved):

PR #105, opened 2023-02-17
feat!: Log Directive support for traces

### LINKED ISSUE #95: Switch PWG to return a list of logs
# Problem

Currently `Directive::Log` exclusively works with println statements. We directly write the information inside of `LogInfo` to std output. This is not ideal as there are multiple ways that someone way want to access logging information in their Noir circuit (writing to file, log traces in memory, etc.). 

The current design also requires us to pass down a flag during acir gen for determining whether to print logs during tests. More details can be found in this PR (https://github.com/noir-lang/noir/pull/630). 

# Solution

Having `Directive::Log` collect logs into a vector that is returned from the PWG would be the ideal solution. This would then enable whomever is calling `backend.solve` to handle the logs as they see fit. Such as printing to std out, writing to file, or accessing a vector of logs in memory. 

# Alternatives considered

There are not many alternatives if we want to enable devs to access the PWG logs in memory. Also it is a superior design to have whomever is calling `backend.solve` decide how to handle the system logs (nargo, JS packages). 

# Additional context

Relevant feature requested in Noir that requires this change: https://github.com/noir-lang/noir/issues/688

vezenovm:
…from printing when solving directives

# Related issue(s)

Resolves #95 

# Description

`Directive::Log` no longer prints the evaluated log while it is being evaluating during partial witness generation. This enables us to have both trace functionality and move logic for displaying logs to the code that calls the PWG rather than keeping it with solver itself.

## Summary of changes

`solve` under the PWG now takes in a vector of directives. These directives are limited to be `Directive::Log` as it does not make sense to be returning any other directive. The log directive that is returned must have have finalized output or else that means there is a critical bug. 

Several more changes showing how to handle logs correctly will be supplied in `nargo`. 

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/directives.rs
+++ b/acir/src/circuit/directives.rs
@@ -58,7 +58,10 @@ pub enum Directive {
         bits: Vec<Witness>, // control bits of the network which permutes the inputs into its sorted version
         sort_by: Vec<u32>, // specify primary index to sort by, then the secondary,... For instance, if tuple is 2 and sort_by is [1,0], then a=[(a0,b0),..] is sorted by bi and then ai.
     },
-    Log(LogInfo),
+    Log {
+        is_trace: bool, // This field states whether the log should be further manipulated or simply displayed to standard output
+        output_info: LogOutputInfo,
+    },
 }
 
 impl Directive {
@@ -159,17 +162,23 @@ impl Directive {
                     write_u32(&mut writer, *i)?;
                 }
             }
-            Directive::Log(info) => match info {
-                LogInfo::FinalizedOutput(output_string) => {
-                    write_bytes(&mut writer, output_string.as_bytes())?;
-                }
-                LogInfo::WitnessOutput(witnesses) => {
-                    write_u32(&mut writer, witnesses.len() as u32)?;
-                    for w in witnesses {
-                        write_u32(&mut writer, w.witness_index())?;
+            Directive::Log {
+                is_trace,
+                output_info,
+            } => {
+                write_u32(&mut writer, *is_trace as u32)?;
+                match output_info {
+                    LogOutputInfo::FinalizedOutput(output_string) => {
+                        write_bytes(&mut writer, output_string.as_bytes())?;
+                    }
+                    LogOutputInfo::WitnessOutput(witnesses) => {
+                        write_u32(&mut writer, witnesses.len() as u32)?;
+                        for w in witnesses {
+                            write_u32(&mut writer, w.witness_index())?;
+                        }
                     }
                 }
-            },
+            }
         };
 
         Ok(())
@@ -274,11 +283,12 @@ impl Directive {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-// If values are compile time and/or known during
-// evaluation, we can form an output string during ACIR generation.
-// Otherwise, we must store witnesses whose values will
-// be fetched during the PWG stage.
-pub enum LogInfo {
+/// This info is used when solving the initial witness
+/// If values are compile time and/or known during
+/// evaluation, we can form an output string during ACIR generation.
+/// Otherwise, we must store witnesses whose values will
+/// be fetched during the PWG stage.
+pub enum LogOutputInfo {
     FinalizedOutput(String),
     WitnessOutput(Vec<Witness>),
 }
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -172,6 +172,8 @@ impl std::fmt::Display for Opcode {
                     LogOutputInfo::FinalizedOutput(output_string) => {
                         write!(f, "Log: {output_string}, log type {is_trace_display}")
                     }
+                    // Note: This assumes that the witnesses have contiguous indices.
+                    // This may not always be the case however we don't want to print out the full vec.
                     LogOutputInfo::WitnessOutput(witnesses) => write!(
                         f,
                         "Log: _{}..._{}, log type: {is_trace_display}",
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -1,6 +1,6 @@
 use std::io::{Read, Write};
 
-use super::directives::{Directive, LogInfo};
+use super::directives::{Directive, LogOutputInfo};
 use crate::native_types::{Expression, Witness};
 use crate::serialization::{read_n, read_u16, read_u32, write_bytes, write_u16, write_u32};
 use crate::BlackBoxFunc;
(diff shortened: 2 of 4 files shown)


acir/src/circuit/opcodes.rs
```diff
@@ -188,15 +188,23 @@ impl std::fmt::Display for Opcode {
...
-                ),
-            },
+            Opcode::Directive(Directive::Log {
+                is_trace,
+                output_info,
+            }) => {
+                let is_trace_display = if *is_trace { "trace" } else { "println" };
+                match output_info {
+                    LogOutputInfo::FinalizedOutput(output_string) => {
+                        write!(f, "Log: {output_string}, log type {is_trace_display}")
```
jfecher, line 198: Appending `, log type {}` after the log rather than before seems odd to me. I'm not aware of a logging framework that does not format logs with the log type before the log itself: `<log type>: <log>`. It also seems odd to me that a `println` would actually print "println" before every message. Am I reading this wrong? Ideally if we do want to print the log types I'd like logs to be displayed as `trace: my trace string` and `my println string` omitting the println entirely.

acir/src/circuit/directives.rs
```diff
@@ -159,17 +162,23 @@ impl Directive {
...
-                    write_bytes(&mut writer, output_string.as_bytes())?;
-                }
-                LogInfo::WitnessOutput(witnesses) => {
-                    write_u32(&mut writer, witnesses.len() as u32)?;
-                    for w in witnesses {
-                        write_u32(&mut writer, w.witness_index())?;
+            Directive::Log {
+                is_trace,
+                output_info,
+            } => {
```
jfecher, line 168: This formatting seems odd to me. Apparently we never copied over the same `.rustfmt.toml` file to this repo when it was split out of noir.

acir/src/circuit/opcodes.rs
```diff
@@ -166,15 +166,20 @@ impl std::fmt::Display for Opcode {
...
+                    }
+                    LogOutputInfo::WitnessOutput(witnesses) => write!(
+                        f,
+                        "Log: _{}..._{}, log type: {is_trace_display}",
+                        witnesses.first().unwrap().witness_index(),
+                        witnesses.last().unwrap().witness_index(),
+                    ),
```
TomAFrench, line 180: We don't have a guarantee that these witnesses are in order/contiguous, do we?

e.g. for the program
```
fn main(x: pub Field, y: Field) -> [Field; 2];
    let z = 2*y
    let return_value = [z, x];
    std::println(return_value);
    return_value 
```

acir/src/circuit/opcodes.rs
TomAFrench: I think you might be misreading this, this isn't quite the log output itself but the display representation of `Log` opcode. We should then display information on how we're logging the data rather than just printing out the message which will be logged at execution time.

acir/src/circuit/opcodes.rs
vezenovm: ah you are right we technically don't. I don't know how I feel about printing the entire witness vec though. Perhaps it is ok as users will most likely only be using println's while debugging?

guipublic (changes requested):
It looks like you need to persist some state during the solve. I also needed this for solving the new block opcode on my PR #114. The solve is no more recursive and we can instantiate data that persist throughout the solve process without adding specific parameters and return values. I suggest you branch from it.

acir/src/circuit/opcodes.rs
TomAFrench: Yeah, I'm not suggesting that we print out a full list as that would get messy for large arrays/strings fast. I'd be fine with just documenting in the code that this output is a small lie.

acvm/src/pwg/directives.rs
```diff
@@ -186,23 +186,27 @@ pub fn solve_directives(
...
+            let witnesses = match output_info {
+                LogOutputInfo::FinalizedOutput(_) => {
+                    return Ok(Some(directive.clone()));
                 }
-                LogInfo::WitnessOutput(witnesses) => witnesses,
+                LogOutputInfo::WitnessOutput(witnesses) => witnesses,
             };
 
             if witnesses.len() == 1 {
+                dbg!(witnesses.clone());
```
TomAFrench, line 200:
```suggestion
```

TomAFrench (approved):
I'm happy with this PR now. @guipublic Can you remove the block as we've addressed the recursive solve function?

acir/src/circuit/directives.rs
```diff
@@ -58,7 +58,10 @@ pub enum Directive {
         bits: Vec<Witness>, // control bits of the network which permutes the inputs into its sorted version
         sort_by: Vec<u32>, // specify primary index to sort by, then the secondary,... For instance, if tuple is 2 and sort_by is [1,0], then a=[(a0,b0),..] is sorted by bi and then ai.
     },
-    Log(LogInfo),
+    Log {
+        is_trace: bool, // This field states whether the log should be further manipulated or simply displayed to standard output
```
guipublic, line 62: Why do we have two separate process, and not just one which basically notifies when the 'log witness' are solved? Then the caller can decide to print to standard output or to further manipulate them. This logic should not be in ACIR.

acvm/src/pwg/directives.rs
```diff
@@ -15,13 +15,13 @@ use super::{get_value, insert_value, sorting::route, witness_to_value};
 pub fn solve_directives(
     initial_witness: &mut BTreeMap<Witness, FieldElement>,
     directive: &Directive,
-) -> Result<(), OpcodeResolutionError> {
+) -> Result<Option<Directive>, OpcodeResolutionError> {
```
guipublic, line 18: I don't think the solve should return an optional directive. The directive is either solved or not. If it is solved, then you can get the witness values from the witness map.

guipublic (changes requested):
My two comments are to remove the logging logic from ACIR and to keep the solving coherent with the other gates.

kevaundray:
> It looks like you need to persist some state during the solve. I also needed this for solving the new block opcode on my PR #114. The solve is no more recursive and we can instantiate data that persist throughout the solve process without adding specific parameters and return values. I suggest you branch from it.

This should not have been a blocking change to this PR. In general, we should not branch off of or copy unrelated PRs. If we want to remove recursion, so that two separate PRs can benefit from it, we should have a PR that does this, merge it into master and then have both PRs rebase.

This is beneficial for two reasons:

- Avoids the situation where two PRs are doing the same thing on two separate branches which will show a diff for the reviewer on both PRs until one of them is merged
- Doesn't block the core of the original PRs which in most cases is unrelated

acvm/src/pwg/directives.rs
kevaundray: Agreed - If a directive is not solved, then it means we did not find a witness that it needed, this will return an Error

vezenovm: I was unsure on whether I should do this but I went with returning a directive mainly to remove duplication of code as the logging information needed by whomever is calling solve is very similar to the Log directive itself. Whomever is calling solve must check that we have a finalized string output and whether the log is a trace or a println. This can be seen in this draft PR: https://github.com/noir-lang/noir/pull/872. I will look into change this as I agree it confuses the idea of whether the directive has been solved or not.

acir/src/circuit/directives.rs
guipublic: I am not suggesting to remove the flag, sorry for the mi-understanding, the location of my comment is confusing. This is because my first reaction was to ask for a clarification on the flag meaning. But at the end, I think it is better to not handle it in the solve, and just solve the witness.
With the flag (from the directive), the caller should be able to handle it correctly, shouldn't he?

vezenovm: Yes that is the case. We only pass the flag along with the solved witness info so that the caller can determine how they should handle the solved log directive. The process in `solve_directives` for `Directive::Log` remains the same for both traces and printlns.

vezenovm: So currently we only support logs of individual fields, arrays, or strings (formatting and handling of complex types can come in a separate PR). In the case of one witness we don't have to construct a list of hex elements. We can simply format the respective witness value and return that the log has been solved. In the case of multiple witnesses this means we have an array, where we have to fetch a list of hex elements, and format them into one finalized string output. Both cases will form a `LogOutputInfo::FinalizedOutput(String)`, which the caller of `solve` will then handle. In the case of a string this is formatted during acir gen and should already be a `LogOutputInfo::FinalizedOutput`. If `Directive::Log` already has finalized output we can just return the string stored in the enum.

guipublic: Thank you for the explanation. For me you are processing the logs here and you should not. The Log directive should just hold the parameters (for now just the flag), and the witness to log.
When the gate is solved, i.e when all the witness are known, you can send the parameters and witness values to the caller. That's it. How/where to print, which language to use, etc... all this should be handled by the caller.

vezenovm: Got it, I see now. Ok I will refactor this

vezenovm:
Closing in favor of PR #137

PR #106, opened 2023-02-20
chore!: make BlackBoxFunc non-exhaustive

TomAFrench:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

Currently adding a new black box function is a breaking change for all backends. For instance in this function we explicitly enumerate all the bb funcs which we do and don't support.

https://github.com/noir-lang/aztec_backend/blob/cff757dca7971161e4bd25e7a744d910c37c22be/barretenberg_static_lib/src/acvm_interop/proof_system.rs#L61-L76

If we were to add a new black box function, rather than defaulting to not supporting it and falling back to arithmetic gates, `aztec_backend` will fail to compile.

This PR changes `enum BlackBoxFunc` to be non-exhaustive so backends must explicitly handle any unknown black box functions when matching on them. This should held reduce the amount of breaking changes related to adding new bb functions going forwards.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/black_box_functions.rs
+++ b/acir/src/circuit/black_box_functions.rs
@@ -2,6 +2,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(test)]
 use strum_macros::EnumIter;
 
+#[non_exhaustive]
 #[allow(clippy::upper_case_acronyms)]
 #[derive(Clone, Debug, Hash, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[cfg_attr(test, derive(EnumIter))]


kevaundray:
In the example you linked, the backend could use `_` to catch all other cases. 

I believe I made it like that in the backend on purpose so that any and all changes to acvm will be flagged on the backend and addressed.

TomAFrench:
Yeah, the idea is that we'd be forcing the backends to include a `_` case.

I don't think leaving it as is is a huge issue for now. My understanding is that in future we're expecting a proliferation of black box functions (due to sumcheck stuff) so breaking every backend each time may be too much trouble.

kevaundray:
Yeah makes sense -- I'd prefer for the backends to explicitly opt into ignoring new blackbox functions instead of us specifying that they need to include `_`

TomAFrench:
Cool, gonna close this PR in that case.

PR #113, opened 2023-02-22
feat(acvm): add `eth_contract_from_vk` to `SmartContract

### LINKED ISSUE #102: Switch SmartContract to accept a verification key instead of a constraint system
# Problem

The `SmartContract` interface is outdated and should reflect the new preprocess method that enables creation of proving and verification keys. We are currently passing a constraint system to the smart contract method which requires computing the proving and verification keys before write a Solidity contract. The need for this has already been highlighted in the [code](https://github.com/noir-lang/acvm/blob/fc3d80b7defedc5f52ffd15d2083bcac2624c851/acvm/src/lib.rs#L124), but making this issue to track.

# Solution

Switch the smart contract method to only accept a byte array representing a verification key rather than a `Circuit`. The C++ method in barretenberg should then be altered to read in a verification key and use it for outputting the solidity contract. The existing C++ method for verifying a proof can be used as reference for this task.

# Alternatives considered

This reduces massive overhead for solidity contract generation and is the best way forward.

# Additional context

(If applicable.)

TomAFrench:
chore: deprecate `eth_contract_from_cs`

# Related issue(s)

Resolves #102 

# Description

## Summary of changes

`eth_contract_from_cs` is now deprecated and `eth_contract_from_vk` has been added which takes a `&[u8]` rather than a `Circuit`.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -115,6 +115,8 @@ pub trait PartialWitnessGenerator {
 }
 
 pub trait SmartContract {
+    // TODO: Allow a backend to support multiple smart contract platforms
+
     /// Takes an ACIR circuit, the number of witnesses and the number of public inputs
     /// Then returns an Ethereum smart contract
     ///
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -115,13 +115,6 @@ pub trait PartialWitnessGenerator {
 }
 
 pub trait SmartContract {
-    // Takes a verification  key and produces a smart contract
-    // The platform indicator allows a backend to support multiple smart contract platforms
-    //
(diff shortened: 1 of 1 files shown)


phated:
A deprecation shouldn't be breaking. It's just a warning that something will have a breaking change in the future.

vezenovm (approved):

TomAFrench:
> Adding a deprecation shouldn't be breaking. It's just a warning that something will have a breaking change in the future.

Whoops, got a bit trigger happy with all the breaking changes recently :sweat_smile:

PR #114, opened 2023-02-22
feat!: add block opcode

guipublic:
# Description

## Summary of changes

This PR adds an opcode which abstract memory operations as read and write to blocks.
It is required for the dynamic array feature because the solver is not able to solve dynamic reads except for simple cases.
The opcode uses the memory operation of a block to emulate the block value during solving in order to resolve the read operations.

Solving does not return an error anymore when it cannot solve a gate, the backend should be adapted to the new solve signature.

# Checklist

# Additional

We will be able to use this new opcode for interfacing with barretenberg when it will expose dynamic memory.


acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    // Try to solve block operations from the trace
+    // As long as operations are resolved, we update/read from the block_value
+    // We stop when an operation cannot be resolved
+    pub fn solve(
+        &mut self,
+        initial_witness: &mut BTreeMap<Witness, FieldElement>,
+        trace: &Vec<BlockOp>,
+    ) -> Result<GateResolution, OpcodeResolutionError> {
```
TomAFrench, line 71: I think this function generally suffers from a tower of if-lets which hampers readability. If we signalled that we should skip this opcode through an `OpcodeResolutionError` rather than `GateResolution`, we could cut this down using `.ok_or_else()?`

acir/src/circuit/opcodes.rs
```diff
@@ -6,11 +6,24 @@ use crate::serialization::{read_n, read_u16, read_u32, write_bytes, write_u16, w
 use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
+pub struct BlockId(u32);
+
+/// Operation on a block
+/// We can either write or read at a block index
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct BlockOp {
```
jfecher, line 15: Do these blocks in acvm have any correspondence to ssa blocks? It seems they are meant to be similar to the MemOps in ssa, I think we should give it a similar name in that case.

acir/src/native_types/arithmetic.rs
```diff
@@ -149,11 +149,23 @@ impl Expression {
...
     }
 
+    // Returns a witness belonging to the expression
+    pub fn get_witness(&self) -> Option<Witness> {
+        if self.linear_combinations.is_empty() {
+            if self.mul_terms.is_empty() {
+                None
+            } else {
+                Some(self.mul_terms[0].1)
+            }
```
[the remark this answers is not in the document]
jfecher: Here's how it would look with both for comparison
```rs
self.linear_combinations.get(0).map(|(_, wit)| *wit)
    .unwrap_or_else(|| self.mul_terms.get(0).map(|(_, wit, _)| *wit))
```

acvm/src/lib.rs
```diff
@@ -58,42 +65,38 @@ pub trait PartialWitnessGenerator {
...
-        }
-        let mut unsolved_opcodes: Vec<Opcode> = Vec::new();
-
-        for opcode in opcodes.into_iter() {
-            let resolution = match &opcode {
-                Opcode::Arithmetic(expr) => ArithmeticSolver::solve(initial_witness, expr),
-                Opcode::BlackBoxFuncCall(bb_func) => {
-                    Self::solve_black_box_function_call(initial_witness, bb_func)
+        let mut unresolved_gates: Vec<Opcode> = Vec::new();
+        let mut blocker = Blocks::default();
```
jfecher, line 71: `blocker` reads like a blocking operation to me. Perhaps we should rename this to `blocks` or similar

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,187 @@
...
+                unknown = op_expr.get_witness();
+                break;
+            }
+            self.solved_operations += 1;
+        }
+        if self.solved_operations == trace.len() {
+            Ok(GateResolution::Resolved)
+        } else {
+            Ok(GateResolution::Skip(OpcodeNotSolvable::MissingAssignment(unknown.unwrap().0)))
+        }
```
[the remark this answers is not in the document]
jfecher: Agreed. Perhaps we can use `ok_or` or similar to reduce the indentation and get the error cases out of the way first as well.

acir/src/native_types/arithmetic.rs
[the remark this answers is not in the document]
guipublic: Yes, I don't see the point to make something clear into something unreadable. Is it more efficient?

acvm/src/pwg/block.rs
guipublic: If there is an error while we solve a gate, then we return the error, but not being able to solve a gate is not an error.

acir/src/circuit/opcodes.rs
guipublic: Indeed, they have nothing to do with ssa blocks, they represent an abstract block of data where you can do read and write operations on it, aka the MemOps in ssa. I am open to change the naming.

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,187 @@
...
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        let mut unknown = None;
+        while self.solved_operations < trace.len() {
+            let op_idx = self.solved_operations;
+            let op_expr = ArithmeticSolver::evaluate(&trace[op_idx].operation, initial_witness);
+            if let Some(operation) = expression_to_const(&op_expr) {
+                let index_expr = ArithmeticSolver::evaluate(&trace[op_idx].index, initial_witness);
+                if let Some(index) = expression_to_const(&index_expr) {
+                    let index = index.try_to_u64().unwrap() as u32;
+                    if operation == FieldElement::zero() {
```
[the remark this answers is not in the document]
guipublic: Done

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,187 @@
...
+    // As long as operations are resolved, we update/read from the block_value
+    // We stop when an operation cannot be resolved
+    pub fn solve(
+        &mut self,
+        initial_witness: &mut BTreeMap<Witness, FieldElement>,
+        trace: &Vec<BlockOp>,
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        let mut unknown = None;
+        while self.solved_operations < trace.len() {
+            let op_idx = self.solved_operations;
```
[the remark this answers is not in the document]
guipublic: Done

acvm/src/pwg/block.rs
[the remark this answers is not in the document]
guipublic: I have changed to return the Skip instead of breaks.
I am not sure what is the suggestion for using ok_or.

acvm/src/lib.rs

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,187 @@
...
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        let solver = self.blocks.entry(id).or_default();
+        solver.solve(solved_witness, trace)
+    }
+}
+
+#[derive(Default)]
+struct BlockSolver {
+    block_value: HashMap<u32, FieldElement>,
+    solved_operations: usize,
```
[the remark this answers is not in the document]
guipublic: Done

acvm/src/pwg/block.rs
TomAFrench: Attempting to solve a gate and finding it to be unsolvable should be an error imo. It's a recoverable error as we may be able to solve it later but it's an error all the same.

guipublic: There is an important distinction between what you call a recoverable error and a definitive error. I explicit this distinction by using an additional type.

TomAFrench (dismissed):
The addition of GateResolution seems unrelated to the purpose of this PR. I think it would be best to split off that change to be reviewed/merged separately.

acir/src/native_types/arithmetic.rs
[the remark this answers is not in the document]
TomAFrench: I find my version more readable tbh, "I want the first thing in this array and then pull out the witness, otherwise None" Vs "is this thing empty? If so then return None otherwise pull out the witness from first element"

In the first you can just read through it whereas I need to scan up and down for the second.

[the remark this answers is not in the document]
guipublic: This is just a personal opinion. 
I am not saying my version is better, but that I prefer mine (because I don't have to read an inner lambda function and try to figure out what this code is really doing and how), and you prefer yours because of the reasons you said.

acir/src/circuit/opcodes.rs
guipublic: I have renamed it to MemOp. At the end (i.e when this PR is merged), ssa will be changed to use this structure instead of the ssa's MemOp.

acvm/src/pwg/block.rs
TomAFrench: I disagree and this is the expected usage of result. For instance if you want to open a file, it may not exist which is recoverable (as you can create the file) or you may not have permissions to open it which is unrecoverable. In both cases this is communicated though an Err though.

https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#matching-on-different-errors

guipublic: This is subject to discussion, the doc also says the following:

> Most languages don’t distinguish between these two kinds of errors and handle both in the same way, using mechanisms such as exceptions. Rust doesn’t have exceptions. Instead, it has the type Result<T, E> for recoverable errors and the panic! macro that stops execution when the program encounters an unrecoverable error. 

So according to the doc we should treat unsolvable as a (recoverable) error, while unsatisfied constraints should panic. That way these two kind of errors are indeed well separated.

acir/src/circuit/opcodes.rs
```diff
@@ -6,11 +6,24 @@ use crate::serialization::{read_n, read_u16, read_u32, write_bytes, write_u16, w
 use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
+pub struct BlockId(u32);
+
+/// Operation on a block
```
kevaundray, line 12: what is a block?

acir/src/native_types/arithmetic.rs
```diff
@@ -149,11 +149,23 @@ impl Expression {
         Ok(expr)
     }
 
-    // TODO: possibly rename, since linear can have one mul_term
```
kevaundray, line 152: why was this comment removed?

acvm/src/compiler/transformers/fallback.rs
```diff
@@ -22,9 +22,8 @@ impl FallbackTransformer {
 
         for opcode in acir.opcodes {
             let bb_func_call = match &opcode {
-                Opcode::Arithmetic(_) | Opcode::Directive(_) => {
-                    // If it is not a black box function, then it is a directive or
-                    // an arithmetic expression which are always supported
+                Opcode::Arithmetic(_) | Opcode::Directive(_) | Opcode::Block(_, _) => {
```
kevaundray, line 25: It seems we are adding a separate opcode, why can it not fit under the category of the ones we already have?

acvm/src/lib.rs
```diff
@@ -58,45 +65,45 @@ pub trait PartialWitnessGenerator {
     fn solve(
         &self,
         initial_witness: &mut BTreeMap<Witness, FieldElement>,
-        mut opcodes: Vec<Opcode>,
+        mut gates_to_resolve: Vec<Opcode>,
```
kevaundray, line 68: Changes in here seem to be unrelated to adding a block opcode, these should be done in a separate PR so that this PR is only concerned with adding the block opcode

acvm/src/lib.rs
```diff
@@ -49,6 +50,12 @@ pub enum OpcodeResolutionError {
     IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
 }
 
+#[derive(PartialEq, Eq, Debug)]
+pub enum GateResolution {
+    Resolved,                // Gate is solved
```
kevaundray, line 55: I think we previously removed this enum, why is it being added again?

kevaundray (changes requested):
This PR is doing more than just adding a block opcode. These changes should be done in a separate PR and or an issue should be opened up for them.

I also left pending questions

acir/src/native_types/arithmetic.rs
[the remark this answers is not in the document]
kevaundray: Why does this method just take the first element from mul_terms of linear_combinations?

- Is it only called when there is either one mul_term of one linear_combination?
- Is it okay that it is ignoring the coefficient for the witness being returned?

In general, I'd change the naming of this and add a more detailed comment, because get_witness() usually infers that there is a canonical Witness to get. Whereas in general an expression can contain many witnesses

acir/src/native_types/arithmetic.rs
guipublic: because a linear expression does not have a mul_term

acvm/src/pwg/block.rs
kevaundray: I think a better question is; what are the errors that are caused by a bug in this code, and what are the errors that are caused by a bug in user input. I class unrecoverable errors under errors which are caused by the acvm code, ie bugs and recoverable errors which are caused by the user input.

Since acvm cannot be sure that the user passed in a satisfiable witness assignment, these are classed as recoverable errors, so we return an error.
Similarly, since acvm cannot be sure that the user passed in an acvm circuit which is "correct" we also see this as a recoverable error and do not panic.

If ACVM was to check the circuit beforehand or have an API that only allowed one to build correct circuits, then it would be okay to panic everywhere else, since we can then assume that the user used our API and then comment that "This method will panic, if the circuit is not correct, which can be the case if you did not use the provided API to build ACIR"

So I guess the TLDR is that these are all users errors and it is upto Noir/frontend to panic since only on the Noir level do we know that this is actually a bug in the compiler.

acir/src/native_types/arithmetic.rs
kevaundray: Agreed, this is my fault for not opening an issue or writing more detail

acir/src/native_types/arithmetic.rs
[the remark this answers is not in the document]
jfecher: > because I don't have to read an inner lambda function and try to figure out what this code is really doing and how

I agree that it is personal preference, though it seems odd to me to avoid lambda functions because you need to read their blocks anymore than you should avoid if statements because you need to read their blocks.

acvm/src/pwg/block.rs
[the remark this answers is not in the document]
jfecher: >  I am not sure what is the suggestion for using ok_or.

For example, on line 77 we currently have:
```rs
if let Some(operation) = expression_to_const(&op_expr) {
```
And to see the else case we must jump all the way down to line 131:
```rs
} else {
    return Ok(GateResolution::Skip(OpcodeNotSolvable::MissingAssignment(
        op_expr.get_witness().unwrap().0,
    )));
}
```
At which point we've likely forgotten what the condition was in the first place. To fix this we can swap the ordering of the two cases, and reduce some rightward shift in the process using `Option::ok_or_else`:
```rs
let operation = expression_to_const(&op_expr).ok_or_else(|| {
    OpcodeNotSolvable::MissingAssignment(
        op_expr.get_witness().unwrap().0,
    ))
})?;
```
Note that `OpcodeNotSolvable` would now need to be either convertable to the error type OpcodeResolutionError, or we'd need to return something else that was. This is the same as Tom's suggestion in the next conversation as well, so feel free to close this one as it is now a duplicate.

acvm/src/pwg/block.rs
guipublic: I agree to treat user errors as recoverable errors, using Result<,> mechanism, and to panic on compiler errors.
What we disagree on is that I think not solving a gate is not an error but a normal situation. It is definitively NOT a user error and also NOT a compiler error. What should be an error is when we cannot solve a circuit, which could be because of a user error (bad input) or an unknown bug.

Thank you @jfecher for the code snippet, we agreed to move this error handling out-of this PR, but I take note of your comments for the next PR.

acir/src/circuit/opcodes.rs
guipublic: I added some comments

acvm/src/compiler/transformers/fallback.rs
guipublic: This one is for emulating memory access. The idea is that a backend could implement it more efficiently.

acvm/src/lib.rs
guipublic: Yes, @TomAFrench already suggested this. I will remove this part and create another PR.
For the record, I copied the non-recursive solved from a long-standing PR which included it, but I should not have include this refactoring.

acvm/src/lib.rs
guipublic: Yes, @TomAFrench already suggested this. I will remove this part and create another PR.
For the record, I copied the non-recursive solved from a long-standing PR which included it, but I should not have include this refactoring.

acvm/src/lib.rs
guipublic: The non-recursive solver is required for solving the block opcode, so this is related. I could/should have done it in a separate PR but I did not anticipated this, it is only at the end of my implementation that I realised that I needed it. At that time I just wanted to commit the PR. But yes, I understand your point.

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,193 @@
...
+use crate::{GateResolution, OpcodeNotSolvable, OpcodeResolutionError};
+
+use super::{
+    arithmetic::{ArithmeticSolver, GateStatus},
+    directives::insert_witness,
+    expression_to_const,
+};
+
+#[derive(Default)]
+pub struct Blocks {
```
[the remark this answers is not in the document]
guipublic: Done

acvm/src/lib.rs
guipublic: It's removed

acvm/src/pwg/block.rs
```diff
@@ -67,69 +71,66 @@ impl BlockSolver {
...
+                        OpcodeNotSolvable::MissingAssignment(value_witness.unwrap().0),
+                    ));
                 }
+            } else if value.is_const() {
+                self.insert_value(index, value.q_c)?;
             } else {
-                unknown = op_expr.get_witness();
-                break;
+                return Err(OpcodeResolutionError::OpcodeNotSolvable(
+                    OpcodeNotSolvable::MissingAssignment(value_witness.unwrap().0),
```
jfecher, line 128: We can factor out the repeated errors into a function:
```rs
let missing_assignment = |witness| 
    OpcodeResolutionError::OpcodeNotSolvable(OpcodeNotSolvable::MissingAssignment(witness));
```

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,188 @@
...
+                        OpcodeNotSolvable::MissingAssignment(value_witness.unwrap().0),
+                    ));
+                }
+            } else if value.is_const() {
+                self.insert_value(index, value.q_c)?;
+            } else {
+                return Err(OpcodeResolutionError::OpcodeNotSolvable(
+                    OpcodeNotSolvable::MissingAssignment(value_witness.unwrap().0),
+                ));
+            }
```
TomAFrench, line 130:
```suggestion
            let value_witness_index = value.any_witness().unwrap().0;
            if value.is_const() {
                self.insert_value(index, value.q_c)?;
            } else if operation.is_zero() && value.is_linear() {
                match ArithmeticSolver::solve_fan_in_term(&value, initial_witness) {
                    GateStatus::GateUnsolvable => {
                        return Err(OpcodeResolutionError::OpcodeNotSolvable(
                            OpcodeNotSolvable::MissingAssignment(value_witness_index),
                        ))
                    }
                    GateStatus::GateSolvable(sum, (coef, w)) => {
                        let map_value = self.get_value(index).ok_or(
                            OpcodeResolutionError::OpcodeNotSolvable(
                                OpcodeNotSolvable::MissingAssignment(w.0),
                            ),
                        )?;
                        insert_witness(w, (map_value - sum - value.q_c) / coef, initial_witness)?;
                    }
                    GateStatus::GateSatisfied(sum) => {
                        self.insert_value(index, sum + value.q_c)?;
                    }
                }
            } else {
                return Err(OpcodeResolutionError::OpcodeNotSolvable(
                    OpcodeNotSolvable::MissingAssignment(value_witness_index),
                ));
            };
```

Couple of changes here:
- `value_witness` always gets unwrapped into its index so we can do this in one place (It would probably be cleaner to convert from `Witness` to index inside https://github.com/noir-lang/acvm/pull/114/files#r1117185097 but we should unwrap early if we know that this witness will always exist.)
- We're recalculating `value` here for no reason I can see.
- branching behaviour always results in the same outcome if `value.is_const()` so pull this to outside of the nested if-statements. This actually allows us to remove nesting altogether.

acvm/src/pwg/block.rs
guipublic: Done

acvm/src/pwg/block.rs
guipublic: Yes good points, thank you for the suggestion. It's updated.

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,164 @@
...
+    // As long as operations are resolved, we update/read from the block_value
+    // We stop when an operation cannot be resolved
+    pub fn solve(
+        &mut self,
+        initial_witness: &mut BTreeMap<Witness, FieldElement>,
+        trace: &[MemOp],
+    ) -> Result<(), OpcodeResolutionError> {
+        let missing_assignment = |witness| {
+            OpcodeResolutionError::OpcodeNotSolvable(OpcodeNotSolvable::MissingAssignment(witness))
+        };
```
TomAFrench, line 78:
```suggestion
        let missing_assignment = |witness| {
            OpcodeResolutionError::OpcodeNotSolvable(OpcodeNotSolvable::MissingAssignment(witness.0))
        };
```

We can let this closure do the final unwrap from witness to index.

guipublic: I also added the first unwrap()

jfecher (approved):

acir/src/native_types/arithmetic.rs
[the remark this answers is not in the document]
TomAFrench: Closing as it's mostly a matter of taste.

acvm/src/pwg/block.rs
TomAFrench: > What we disagree on is that I think not solving a gate is not an error but a normal situation.

My view on this is that it's definitely an error on the scale of these lower level functions (it's been provided inputs which it is unable to process) but it isn't an error on the compiler level (which is why we catch them there).

I'm going to resolve this comment and we can address `GateResolution` vs errors later.

acir/src/circuit/opcodes.rs
```diff
@@ -6,11 +6,25 @@ use crate::serialization::{read_n, read_u16, read_u32, write_bytes, write_u16, w
...
+/// We can either write or read at a block index
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MemOp {
+    pub operation: Expression,
```
TomAFrench, line 16: As this mirrors the linked review comment in `noir-lang/noir`, I think we should resolve this before merging to ensure that we're treating this consistently.

https://github.com/noir-lang/noir/pull/886/files/2aa2fb5fe16863ca7fabb269dbef601adf9585fa#r1114772704

TomAFrench (commented):
Should be gtg after this.

acvm/src/pwg/block.rs
```diff
@@ -0,0 +1,164 @@
...
+            std::collections::hash_map::Entry::Vacant(e) => {
+                e.insert(value);
+            }
+            std::collections::hash_map::Entry::Occupied(e) => {
+                if e.get() != &value {
+                    return Err(OpcodeResolutionError::UnsatisfiedConstrain);
+                }
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
guipublic: I don't think it would be faster but I do agree it is more readable.

acir/src/circuit/opcodes.rs
guipublic: I answered in the PR and resolved the conversation, so I will resolve this one also.

jfecher (approved):

acir/src/native_types/arithmetic.rs
```diff
@@ -149,6 +149,20 @@ impl Expression {
         Ok(expr)
     }
 
+    // Returns one witness belonging to the expression, in no relevant order
+    // Returns None if the expression is const
+    pub fn any_witness(&self) -> Option<Witness> {
```
kevaundray, line 154: Can you put a brief comment on why this is needed? It seems like its just for when we return Errors?

Not blocking for this PR, but I think we can refactor it later so that its not in the namespace of arithmetic and is instead a free-standing function in the module that needs it

guipublic: It is not needed strictly speaking, I use it when evaluating an expression during solving to report one non-solved witness. It could be move to the pwg part. I have added a comment on its usage.

acvm/src/pwg.rs
```diff
@@ -7,21 +7,33 @@ use acir::{
...
+    witness: Witness,
+) -> Option<&FieldElement> {
+    initial_witness.get(&witness)
+}
+
 // Returns the concrete value for a particular witness
 // If the witness has no assignment, then
 // an error is returned
-pub fn witness_to_value(
+pub fn witness_to_value_unwrap(
```
kevaundray, line 36: prefixing a function name with unwrap has a particular meaning in Rust; reading the function name gives me the impression that inside it will be calling .unwrap and panicking. 

Why can't you use witness_to_value as it was when it was returning a Result, instead of creating this new function and having a different function that returns Option?

acvm/src/pwg.rs
```diff
@@ -30,37 +42,31 @@ pub fn witness_to_value(
         None => Err(OpcodeNotSolvable::MissingAssignment(witness.0).into()),
     }
 }
+
+pub fn expression_to_const(expr: &Expression) -> Option<FieldElement> {
```
kevaundray, line 46: Lets put this as a method on `Expression`:
```rust
impl Expression{

   pub fn to_const(&self) -> Option<FieldElement>{
      self.is_const().then_some(self.q_c)
   } 

}
```

acvm/src/pwg.rs
```diff
@@ -30,37 +42,31 @@ pub fn witness_to_value(
...
+) -> Option<FieldElement> {
+    expression_to_const(&ArithmeticSolver::evaluate(expr, initial_witness))
+}
 
-        result += coefficient * lhs_assignment * rhs_assignment;
+pub fn get_value_unwrap(
+    expr: &Expression,
+    initial_witness: &BTreeMap<Witness, FieldElement>,
+) -> Result<FieldElement, OpcodeResolutionError> {
+    let expr = ArithmeticSolver::evaluate(expr, initial_witness);
```
kevaundray, line 63: same comment about `_unwrap`, I'm not sure why we need these instead of just using the original get_value that returned a Result

acir/src/native_types/arithmetic.rs
kevaundray: A comment is fine, would be better to move it to where its needed

acvm/src/pwg.rs
guipublic: Sure, it's done.

acvm/src/pwg.rs
guipublic: Done

acvm/src/pwg.rs
guipublic: I guess it made more sens with the GateResolution change, I don't remember.
I have restored the previous name and removed the other.

acir/src/native_types/arithmetic.rs
guipublic: I can move it to pwg

guipublic: Done

kevaundray (approved):
LGTM

PR #100, opened 2023-02-16
chore: Fix small nits for ToRadix

kevaundray:
# Related issue(s)

(If it does not already exist, first create a GitHub issue that describes the problem this Pull Request (PR) solves before creating the PR and link it here.)

Resolves (link to issue)

# Description

This does a small refactor of the ToRadix directive. Since both little endian and big endian rely on the number of witnesses being assigned to, to be more than the number of decomposed digits, this was added for both branches.

We also now pad the decomposed integer to the beginning or end depending on its endianness.

## Summary of changes

(Describe the changes in this PR. Point out breaking changes if any.)

## Dependency additions / changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` with default settings.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this PR to the issue(s) that it resolves.
- [ ] I have reviewed the changes on GitHub, line by line.
- [ ] I have ensured all changes are covered in the description.

# Additional context

(If applicable.)

PR #129, opened 2023-03-01
chore!(acir): remove is_little_endian from to_radix

guipublic:
# Description

## Summary of changes

In order to keep ACIR simple, I remove the endianness handling so that it only support to_radix_le. The gate as been renamed accordingly.

# Checklist

### DIFF
--- a/acir/src/circuit/directives.rs
+++ b/acir/src/circuit/directives.rs
@@ -43,7 +43,7 @@ pub enum Directive {
     },
 
     //decomposition of a: a=\sum b[i]*radix^i where b is an array of witnesses < radix in little endian form
-    ToRadixLe {
+    ToLeRadix {
         a: Expression,
         b: Vec<Witness>,
         radix: u32,
@@ -67,7 +67,7 @@ impl Directive {
             Directive::Quotient { .. } => "quotient",
             Directive::Truncate { .. } => "truncate",
             Directive::OddRange { .. } => "odd_range",
-            Directive::ToRadixLe { .. } => "to_radix_le",
+            Directive::ToLeRadix { .. } => "to_le_radix",
             Directive::PermutationSort { .. } => "permutation_sort",
             Directive::Log { .. } => "log",
         }
@@ -78,7 +78,7 @@ impl Directive {
             Directive::Quotient { .. } => 1,
             Directive::Truncate { .. } => 2,
             Directive::OddRange { .. } => 3,
-            Directive::ToRadixLe { .. } => 4,
+            Directive::ToLeRadix { .. } => 4,
             Directive::Log { .. } => 5,
             Directive::PermutationSort { .. } => 6,
         }
@@ -116,7 +116,7 @@ impl Directive {
                 write_u32(&mut writer, r.witness_index())?;
                 write_u32(&mut writer, *bit_size)?;
             }
-            Directive::ToRadixLe { a, b, radix } => {
+            Directive::ToLeRadix { a, b, radix } => {
                 a.write(&mut writer)?;
                 write_u32(&mut writer, b.len() as u32)?;
                 for bit in b {
@@ -206,7 +206,7 @@ impl Directive {
 
                 let radix = read_u32(&mut reader)?;
 
-                Ok(Directive::ToRadixLe { a, b, radix })
+                Ok(Directive::ToLeRadix { a, b, radix })
             }
             6 => {
                 let tuple = read_u32(&mut reader)?;
@@ -285,14 +285,14 @@ fn serialization_roundtrip() {
     let odd_range =
         Directive::OddRange { a: Witness(1u32), b: Witness(2u32), r: Witness(3u32), bit_size: 32 };
 
-    let to_radix_le = Directive::ToRadixLe {
+    let to_le_radix = Directive::ToLeRadix {
         a: Expression::default(),
         b: vec![Witness(1u32), Witness(2u32), Witness(3u32), Witness(4u32)],
         radix: 4,
     };
 
     let directives =
-        vec![invert, quotient_none, quotient_predicate, truncate, odd_range, to_radix_le];
+        vec![invert, quotient_none, quotient_predicate, truncate, odd_range, to_le_radix];
 
     for directive in directives {
         let (dir, got_dir) = read_write(directive);
--- a/acir/src/circuit/directives.rs
+++ b/acir/src/circuit/directives.rs
@@ -42,12 +42,11 @@ pub enum Directive {
         bit_size: u32,
     },
 
-    //decomposition of a: a=\sum b[i]*radix^i where b is an array of witnesses < radix in either little endian or big endian form
-    ToRadix {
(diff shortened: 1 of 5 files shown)


jfecher:
We no longer want to support converting to big endian at all?

kevaundray:
The rationale for this is so that upstream users don't need to do it themselves. It's not that complex imo and if it was then that would be an argument to put it here instead of having multiple upstream users do it themselves.

Btw, was there a reason you chosen ToRadixLE instead of ToRadixBE?

kevaundray:
> We no longer want to support converting to big endian at all?

It would need to be handled by upstream users -- which is why I'm not particularly in favour

guipublic:
> Btw, was there a reason you chosen ToRadixLE instead of ToRadixBE?

No, but it was just the solution with the less impact.

> The rationale for this is so that upstream users don't need to do it themselves.

I strongly disagree here, I don't know what you call upstream users but for acir backends, keeping acid simple and lean is definitively the best.

kevaundray:
> > Btw, was there a reason you chosen ToRadixLE instead of ToRadixBE?
> 
> 
> 
> No, but it was just the solution with the less impact.
> 
> 
> 
> > The rationale for this is so that upstream users don't need to do it themselves.
> 
> 
> 
> I strongly disagree here, I don't know what you call upstream users but for acir backends, keeping acir simple and lean is definitively the best.

- How are backends affected by directives?

- Do you mean to keep the external API simple or the internals of ACIR simple?

guipublic:
> * How are backends affected by directives?
> * Do you mean to keep the external API simple or the internals of ACIR simple?

Backends are not affected by directives, but users generating ACIR are affected as they need to provide correct inputs to the directives. What I mean is that the ACIR opcodes should remain simple and low-level so that backends can easily integrate with. Directive are part of ACIR opcodes and should follow the same principles.
Internally, ACIR can implement complex stuff to provide useful features. For example a powerful solver allows us to limit the usage of directives.

kevaundray:
> > * How are backends affected by directives?
> > * Do you mean to keep the external API simple or the internals of ACIR simple?
> 
> Backends are not affected by directives, but users generating ACIR are affected as they need to provide correct inputs to the directives. What I mean is that the ACIR opcodes should remain simple and low-level so that backends can easily integrate with. Directive are part of ACIR opcodes and should follow the same principles. Internally, ACIR can implement complex stuff to provide useful features. For example a powerful solver allows us to limit the usage of directives.

How does this boolean flag make it anymore complex? @jfecher and @TomAFrench what are your thoughts here? It's not something that I'd die on a hill for, but I don't understand what complexity is being avoided here especially if we are not sure that LittleEndian will be the popular choice.

kevaundray:
> 

Yeah if BigEndian is for some reason more popular then upstream frontends will need to reverse the bytes -- in general ACIR should hold as much "shared" code as possible

jfecher:
I believe the complexity that is being avoided is with this change any future backends of ACIR will no longer need to implement LE and BE conversions, only LE.

kevaundray:
> I believe the complexity that is being avoided is with this change any future backends of ACIR will no longer need to implement LE and BE conversions, only LE.

To be clear, currently backends don't implement directives, so it would be more around frontends like Noir and how they call it. This is different to our other opcodes which are not directives like sha256 which will need to be implemented by backends.

So the complexity we are avoiding here is the extra boolean flag to say whether its little/big endian and the assosciated code _internal_ to ACIR

guipublic:
IMHO it is better to not deal with endianness in ACIR, it is a bit like asking your processor to handle it. This is really a frontend thing.
Handling big endian requires this new unreachable error which is making internal ACIR more complex, but also ACIR interactions with frontends more complex.

kevaundray:
> IMHO it is better to not deal with endianness in ACIR, it is a bit like asking your processor to handle it. This is really a frontend thing.
> Handling big endian requires this new unreachable error which is making internal ACIR more complex, but also ACIR interactions with frontends more complex.

[This PR](https://github.com/noir-lang/acvm/pull/100/files) got rid of the unreachable error and simplified the handling.

kevaundray:
> want to add that if `acir` supports big endian. It seems to me that the front end has to implement constrains for big endian. I didn't realize this until working on the front end part of this.
> 
> It is probably my bad that I didn't notice there is a simpler way to implement big endian in the first place, which result in a lot of efforts reviewing my code. sorry abt this 🥲

Interesting, I'm not sure why since its just a directive. It should only stop the frontend from needing to reverse. @guipublic can you comment here?

@Ethan-000 since you were the steward for this change and I don't feel particularly strongly about it, if you are for it, then we can push it through and then state that all directives need to explicitly specify and be little endian moving forward.

kevaundray:
> > > want to add that if `acir` supports big endian. It seems to me that the front end has to implement constrains for big endian. I didn't realize this until working on the front end part of this.
> > > It is probably my bad that I didn't notice there is a simpler way to implement big endian in the first place, which result in a lot of efforts reviewing my code. sorry abt this 🥲
> > 
> > 
> > Interesting, I'm not sure why since its just a directive. It should only stop the frontend from needing to reverse. @guipublic can you comment here?
> > @Ethan-000 since you were the steward for this change and I don't feel particularly strongly about it, if you are for it, then we can push it through and then state that all directives need to explicitly specify and be little endian moving forward.
> 
> I'm not sure either cus I thought directives doesn't need constrains, but apparently they were implemented here. https://github.com/noir-lang/noir/blob/e51ff01b426c7f91422a53e691b6aaf9bb4b4a86/crates/noirc_evaluator/src/ssa/acir_gen/constraints.rs#L435.
> 
> I think it should be fine to implement big endian by reversing the result otherwise big endian will need similar constrains. I've tried this and got one implementation working but I'm not sure if the implementation makes sense. Implement `ToRadixLe` as opposed to `ToRadixBe` here can lead to easier constrains implementation in the front end I think

Oh I see what you mean, the constraints to ensure that the directive was correct. This wouldn't change with either impl, you can still constrain with little endian

kevaundray:
> I'd propose to change it to `ToLeRadix` and 'to_le_radix' here actually to match Rust.

Alright lets go with that and we can close the nits PR I made -- @guipublic can you also remove the unreachable error variant?

kevaundray:
> > > I'd propose to change it to `ToLeRadix` and 'to_le_radix' here actually to match Rust.
> 
> > 
> 
> > Alright lets go with that and we can close the nits PR I made -- @guipublic can you also remove the unreachable error variant?
> 
> > 
> 
> > EDIT:
> 
> > 
> 
> > You are also proposing to change the naming, I don't have a preference for this, as long as we are consistent
> 
> 
> 
> sorry about the trouble sir 😭. 
> 
> 
> 
> I don't have a naming preference either. It seem some popular crates are not consistent with rust anyways. Either should be okay. I'll change the front end if we decide to go with `to_radix_le`

No problem ser!

guipublic:
I renamed to ToLeRadix and removed the unreachable error.

kevaundray (approved):

PR #132, opened 2023-03-02
fix!(acvm): report an error when solver is stalled

guipublic:
# Description

## Summary of changes

The PR distinguish expected outcomes from solving a gate vs errors happening during solving. This allows us to report an error when the solver is stalled instead of infinite loop/stack overflow.

I believe the backend needs to be updated.

# Checklist

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -35,8 +35,6 @@ pub enum OpcodeNotSolvable {
     MissingAssignment(u32),
     #[error("expression has too many unknowns {0}")]
     ExpressionHasTooManyUnknowns(Expression),
-    #[error("compiler error: unreachable code")]
-    UnreachableCode,
 }
 
 #[derive(PartialEq, Eq, Debug, Error)]
@@ -50,7 +48,13 @@ pub enum OpcodeResolutionError {
     #[error("unexpected opcode, expected {0}, but got {1}")]
     UnexpectedOpcode(&'static str, BlackBoxFunc),
     #[error("expected {0} inputs for function {1}, but got {2}")]
-    IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
+    IncorrectNumFunctionArguments(usize, String, usize),
+}
+
+pub enum GateResolution {
+    Solved,                     // The gate is resolved
+    Stalled(OpcodeNotSolvable), // The gate is not solvable
+    InProgress,                 // The gate is not solvable but could resolved some witness
 }
 
 pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator {}
@@ -68,7 +72,8 @@ pub trait PartialWitnessGenerator {
         let mut blocks = Blocks::default();
         while !opcode_to_solve.is_empty() {
             unresolved_opcodes.clear();
-
+            let mut stalled = true;
+            let mut opcode_not_solvable = None;
             for opcode in &opcode_to_solve {
                 let resolution = match opcode {
                     Opcode::Arithmetic(expr) => ArithmeticSolver::solve(initial_witness, expr),
@@ -81,17 +86,31 @@ pub trait PartialWitnessGenerator {
                     Opcode::Block(id, trace) => blocks.solve(*id, trace, initial_witness),
                 };
                 match resolution {
-                    Ok(_) => {
-                        // We do nothing in the happy case
+                    Ok(GateResolution::Solved) => {
+                        stalled = false;
                     }
-                    Err(OpcodeResolutionError::OpcodeNotSolvable(_)) => {
-                        // For opcode not solvable errors, we push those opcodes to the back as
+                    Ok(GateResolution::InProgress) => {
+                        stalled = false;
+                        unresolved_opcodes.push(opcode.clone());
+                    }
+                    Ok(GateResolution::Stalled(not_solvable)) => {
+                        if opcode_not_solvable.is_none() {
+                            // we keep track of the first unsolvable opcode
+                            opcode_not_solvable = Some(not_solvable);
+                        }
+                        // We push those opcodes not solvable to the back as
                         // it could be because the opcodes are out of order, i.e. this assignment
                         // relies on a later opcodes' results
                         unresolved_opcodes.push(opcode.clone());
                     }
+                    Err(OpcodeResolutionError::OpcodeNotSolvable(_)) => {
+                        unreachable!("ICE - Result should have been converted to GateResolution")
+                    }
                     Err(err) => return Err(err),
                 }
+            }
+            if stalled && !unresolved_opcodes.is_empty() {
+                return Err(OpcodeResolutionError::OpcodeNotSolvable(opcode_not_solvable.unwrap()));
             }
             std::mem::swap(&mut opcode_to_solve, &mut unresolved_opcodes);
         }
@@ -101,7 +120,7 @@ pub trait PartialWitnessGenerator {
     fn solve_black_box_function_call(
         initial_witness: &mut BTreeMap<Witness, FieldElement>,
         func_call: &BlackBoxFuncCall,
-    ) -> Result<(), OpcodeResolutionError>;
+    ) -> Result<GateResolution, OpcodeResolutionError>;
 
     // Check if all of the inputs to the function have assignments
     // Returns true if all of the inputs have been assigned
@@ -117,8 +136,14 @@ pub trait PartialWitnessGenerator {
     fn solve_directives(
         initial_witness: &mut BTreeMap<Witness, FieldElement>,
         directive: &Directive,
-    ) -> Result<(), OpcodeResolutionError> {
-        pwg::directives::solve_directives(initial_witness, directive)
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        match pwg::directives::solve_directives(initial_witness, directive) {
+            Ok(_) => Ok(GateResolution::Solved),
+            Err(OpcodeResolutionError::OpcodeNotSolvable(unsolved)) => {
+                Ok(GateResolution::Stalled(unsolved))
+            }
+            Err(err) => Err(err),
+        }
     }
 }
 
--- a/acvm/src/pwg/block.rs
+++ b/acvm/src/pwg/block.rs
@@ -6,7 +6,7 @@ use acir::{
     FieldElement,
 };
 
-use crate::{OpcodeNotSolvable, OpcodeResolutionError};
+use crate::{GateResolution, OpcodeNotSolvable, OpcodeResolutionError};
 
 use super::{
     arithmetic::{ArithmeticSolver, GateStatus},
@@ -25,7 +25,7 @@ impl Blocks {
         id: BlockId,
         trace: &[MemOp],
         solved_witness: &mut BTreeMap<Witness, FieldElement>,
-    ) -> Result<(), OpcodeResolutionError> {
+    ) -> Result<GateResolution, OpcodeResolutionError> {
         let solver = self.blocks.entry(id).or_default();
         solver.solve(solved_witness, trace)
     }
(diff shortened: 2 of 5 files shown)


acvm/src/pwg/directives.rs
```diff
@@ -103,18 +103,23 @@ pub fn solve_directives(
...
                 let mut value = FieldElement::zero();
                 for (i, witness) in b.iter().enumerate() {
                     if i >= padding_len {
-                        value = match decomposed_integer.get(i - padding_len) {
-                            Some(digit) => FieldElement::from_be_bytes_reduce(&[*digit]),
-                            None => return Err(OpcodeNotSolvable::UnreachableCode.into()),
-                        };
+                        value = FieldElement::from_be_bytes_reduce(&[
+                            decomposed_integer[i - padding_len]
+                        ]);
```
TomAFrench, line 122: Nice, yeah I've been wanting to nix this error as well.

acvm/src/lib.rs
```diff
@@ -117,8 +137,14 @@ pub trait PartialWitnessGenerator {
...
-    ) -> Result<(), OpcodeResolutionError> {
-        pwg::directives::solve_directives(initial_witness, directive)
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        match pwg::directives::solve_directives(initial_witness, directive) {
+            Ok(_) => Ok(GateResolution::Solved),
+            Err(OpcodeResolutionError::OpcodeNotSolvable(unsolved)) => {
+                Ok(GateResolution::Stalled(unsolved))
+            }
+            Err(err) => Err(err),
+        }
```
TomAFrench, line 147: This smells like we're leaking a concept which is only relevant to `blocks.solve()` into the other types of opcode.

TomAFrench (commented):
I think we're working too hard to have a single consistent return type for all the different solvers and this causes details of `BlockSolver` to leak into the other solvers.

My understanding of this is that while we need some way to signal partial progress within a block (as each block "opcode" is made of several arithmetic opcodes), this is not necessary for `Arithmetic`, `BlackBoxFuncCall` or `Directive` as they're all atomic actions.

In order to make this distinction clearer in the code I'd recommend that we restrict the `Stalled/InProgress` return type distinction to when dealing with blocks. We've already said that we should add more structure based on opcode type here: https://github.com/noir-lang/acvm/blob/efb48516f2a6f73225ab3e64d8ebcf4ab6c71a84/acvm/src/lib.rs#L29-L31

We should also document `BlockSolver` on the fact that its non-atomic as it's a special case.

acvm/src/lib.rs
guipublic: This the part where we do the distinction between real errors and GateResolution. `Stalled` and `Solved` are relevant to all opcodes.

acvm/src/pwg/block.rs
```diff
@@ -65,43 +65,60 @@ impl BlockSolver {
...
-                witness.unwrap().0,
-            ))
+    ) -> Result<GateResolution, OpcodeResolutionError> {
+        let mut result = None;
+        let compute_result = |witness: Option<Witness>, result: Option<GateResolution>| {
+            if let Some(result) = result {
+                return Ok(result);
+            }
+            Ok(GateResolution::Stalled(OpcodeNotSolvable::MissingAssignment(witness.unwrap().0)))
         };
```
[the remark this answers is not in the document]
guipublic: Done

acvm/src/lib.rs
```diff
@@ -50,7 +48,14 @@ pub enum OpcodeResolutionError {
     #[error("unexpected opcode, expected {0}, but got {1}")]
     UnexpectedOpcode(&'static str, BlackBoxFunc),
     #[error("expected {0} inputs for function {1}, but got {2}")]
-    IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
+    IncorrectNumFunctionArguments(usize, String, usize),
```
kevaundray, line 51: why did this get changed from BlackBoxFunc to string?

acvm/src/lib.rs
```diff
@@ -50,7 +48,14 @@ pub enum OpcodeResolutionError {
...
+}
+
+#[derive(Debug, PartialEq)]
+pub enum GateResolution {
+    Solved,                     // The gate is resolved
+    Stalled(OpcodeNotSolvable), // The gate is not solvable
+    InProgress,                 // The gate is not solvable but could resolved some witness
```
kevaundray, line 58: Can you explain why these two variants, Stalled and InPogress are needed at this layer? It seems like an abstraction leak from the Blocksolver

acvm/src/lib.rs
```diff
@@ -81,18 +87,32 @@ pub trait PartialWitnessGenerator {
...
                         unresolved_opcodes.push(opcode.clone());
                     }
+                    Err(OpcodeResolutionError::OpcodeNotSolvable(_)) => {
+                        unreachable!("ICE - Result should have been converted to GateResolution")
+                    }
                     Err(err) => return Err(err),
                 }
             }
+            if stalled && !unresolved_opcodes.is_empty() {
+                return Err(OpcodeResolutionError::OpcodeNotSolvable(opcode_not_solvable.unwrap()));
```
kevaundray, line 114: general nit: .unwrap should explain why its okay to unwrap with expect

kevaundray (changes requested):
It seems that implementation details of the BlockSolver has leaked into the top layers. I'd suggest to put GateResolution (stalling and in-progress) into the block solver module and then handle it there.

Can you explain why the block solver needs in progress? Its possible that maybe we need more fine-grained opcodes

acvm/src/lib.rs
guipublic: If you have only gates that are solved/unsolved you cannot determine if you are stalled or not. A gate individually cannot know if the whole process is stalled or not so you cannot confine the enum to the Blocksolver.

acvm/src/lib.rs

acvm/src/lib.rs
guipublic: Because it is also used for the ToRadix directive.

acvm/src/lib.rs
TomAFrench: We have two forms of stalling here however, a `Block` can stall and also the entire solving can stall.

The first we can catch easily within `block.solve` and return like we have in `GateResolution`.

The second is different as it's global. All the other opcode kinds can return pass/fail and still feed into this value just fine. We then just need to handle the `InProgress/Stalled` variants differently for `Block` opcodes when setting the global stalled flag.

TomAFrench: Making all opcodes return a `GateResolution` however creates ambiguity about "can a black box function progress but not be fully solved?" where you need to check all the potential return values.

guipublic: Indeed we check the 3 possible return values, there is no problem with that.

acvm/src/lib.rs
kevaundray: Perhaps we could use Opcode instead of String?

acvm/src/lib.rs
guipublic: Indeed I don't understand what you want to say. If someone is new to the codebase and wants to know if an Arithmetic  pass or fail, then it's easy, either the solve returns an error which means it is failing, or not meaning the pass succeeded. Furthermore, the return value indicates when the gate is solved.

acvm/src/lib.rs
guipublic: ToRadix does not use it anymore, so I reverted to blackboxfunc.

guipublic:
> Can you explain why the block solver needs in progress? Its possible that maybe we need more fine-grained opcodes

It needs to be resolved in order, from the start of the vector, until the end. Fined-grained opcodes would loose this ordering. It is not possible to report stalled/skipped although some progress is done as the solver will falsely believe it is stalled. It is not possible to report solved if only some progress is done as the solver will not try to solve it anymore.

acvm/src/lib.rs
TomAFrench: My concern isn't with seeing if a particular invocation of the solver returned a `Solved` or `Stalled` but rather that the fact that it returns a `GateResolution` implies that it is able to signal that it's made partial progress. In order to verify that the `Arithmetic` solver is "all or nothing" a human needs to verify that it never returns `GateResolution::InProgress`.

My opinion is that we should make a conscious choice about which opcodes are allowed to signal that they can partially progress to provide more information on the execution flow inside the ACVM, i.e. which opcodes return `GateResolution` vs a restricted pass/fail value.

For instance we should ask ourselves whether we want a blackbox function be able to signal partial progress? All the instances we have currently are "all or nothing" so why allow a backend to return `GateResolution::InProgress`?

acvm/src/lib.rs
```diff
@@ -51,6 +51,13 @@ pub enum OpcodeResolutionError {
     IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
 }
 
+#[derive(Debug, PartialEq)]
+pub enum GateResolution {
+    Solved,                     // The gate is resolved
+    Stalled(OpcodeNotSolvable), // The gate is not solvable
+    InProgress,                 // The gate is not solvable but could resolved some witness
```
TomAFrench, line 58:
```suggestion
#[derive(Debug, PartialEq)]
enum AtomicOpcodeResolution {
    /// The opcode is resolved
    Solved,
    /// The opcode is not solvable
    Stalled(OpcodeNotSolvable), 
}

#[derive(Debug, PartialEq)]
pub enum OpcodeResolution {
    /// The opcode is resolved
    Solved,
    /// The opcode is not solvable
    Stalled(OpcodeNotSolvable),
    /// The opcode is not solvable but could resolve some witnesses
    InProgress,                 
}

// TODO implement OpcodeResolution::from<AtomicOpcodeResolution>
```

My thinking is that we'd have something along these lines. `ArithmeticSolver`/`DirectiveSolver` would return an `AtomicOpcodeResolution` and then we convert to an `OpcodeResolution` inside the global `solve` method.

acvm/src/lib.rs
```diff
@@ -51,6 +51,13 @@ pub enum OpcodeResolutionError {
     IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
 }
 
+#[derive(Debug, PartialEq)]
+pub enum GateResolution {
```
TomAFrench, line 55: In any case, this should be renamed to `OpcodeResolution`

acvm/src/lib.rs
TomAFrench: In any case, this should be renamed to `OpcodeResolution` and have the comments changed to doc comments.

acvm/src/lib.rs
```diff
@@ -79,18 +87,35 @@ pub trait PartialWitnessGenerator {
...
+                    Err(OpcodeResolutionError::OpcodeNotSolvable(_)) => {
+                        unreachable!("ICE - Result should have been converted to GateResolution")
+                    }
                     Err(err) => return Err(err),
                 }
             }
+            if stalled && !unresolved_opcodes.is_empty() {
+                return Err(OpcodeResolutionError::OpcodeNotSolvable(
+                    opcode_not_solvable
+                        .expect("infallible: cannot be stalled and none at the same time"),
```
TomAFrench, line 116:
```suggestion
                        .expect("infallible: cannot be stalled and None at the same time"),
```

Nit but this makes it clearer that we're talking about an actual `None` value.

acvm/src/lib.rs
guipublic: Thank you for the clarification.
For me, I don't think we should make this choice at the gate level, it is the responsibility of the solver to decide the gate status. If it can only be `solved` or `stalled`, then  you can see this in the solver.

If for example you consider arithmetic gate, a `binary solver` will be able to make some progress without solving the gate by identifying some witness as boolean. The basic `arithmetic solver` on the other is either stall or solved.

TomAFrench: The core of my point however is that we should codify which solvers are atomic vs non-atomic by their return value, not requiring people so scan through them for all of their potential return values.

kevaundray (approved):

kevaundray:
I think we can address the style issues and refactoring in a later PR. 

For the Block opcode, because of the way it is structured, this is needed. I tried to do a thought experiment with splitting up the block opcode into a MemRead, MemWrite and MemInit, but it surfaced the same problems since writing and reading are dependent on each other. Thanks to @guipublic for walking through this with me.

PR #136, opened 2023-03-03
Joss/stepwise pwg wasm

### LINKED ISSUE #134: Expose ACVM's intermediate witness solving to JS decoupled from backend
### Problem

Capabilities such as witness solving, proving, verifying, compiling circuits, and generating smart contract verifiers are accessed via the `Backend` trait, which is implemented on a per proving system basis. While this has worked in the context of nargo, it's far too restrictive for productions systems which require only certain capabilities in their differing components.

The most pressing concern is that a web browser (user client) needs only the ability to solve the intermediate witness and then feed it into a browser-specialised implementation of a prover. At present the tight coupling of backend capabilities prevents a web browser from loading a prover that's optimised to use multiple web workers, without massive redundancy and possible version mismatches. (Yet to be investigated: I also predict memory management issues.)

### Proposed solution

The first step towards unpicking this tight coupling is to allow the web browser to access ACVM's intermediate witness solving in a manner that is backend agnostic. Then the user client is free to use it in conjunction with a specialised prover implementation of their choosing.

### Alternatives considered

Completely decouple web browser from nargo composition restrictions by reimplementing the required capabilities directly in typescript.

### Additional context

An attempt at visualising the composition, and the direction we should take to improve modularity:

Screenshot 2023-03-03 at 09 27 59

joss-aztec:
# Related issue(s)

Resolves #134

# Description

## Summary of changes

Exposes ACVM's intermediate witness solving capabilities to a JS context in a backend agnostic manner. Blackbox solutions are handled as js callbacks with async support. The structure is a follows:

- The ACVM JS user calls `solveIntermediateWitness` with callbacks for blackbox functions that can't yet be solved deterministically
- ACVM JS asynchronously coordinates the callbacks with a task held ACVM wasm's memory
- ACVM wasm stores and invokes ongoing tasks via a singleton (reducing what could be huge data marshalling costs)
- ACVM implements the task's detail in `StepwisePwgGenerator`, which mirrors the behaviour of the existing `PartialWitnessGenerator`, but differs in that:
  - it progresses to completion via multiple invocations from its parent
  - it directly solves blackboxes for which we already have convenient deterministic implementations
  - it shorts the loop when it encounters a blackbox due for solving externally, and propagates it up

Two new packages dependant on `acvm` crate:
- `acvm_wasm` crate, which implements a singleton task manager
- `acvm_js` node package, which wraps the wasm and drives the task loop
  - Note the build product is named explicitly for the curve that's compiled into the wasm `@noir-lang/acvm-bn254`

# Checklist

# Additional context

This work begins paving the way towards improving modularity of components in the noir ecosystem.


guipublic:
It seems to me that we are trying a solve a BB specific issue that is has with the web by putting in place a complex architecture involving acvm, js and backends.
Isn't the simpler solution to have only deterministic opcodes? 
I believe we are close to ultra-plonk integration, which should give us that, shouldn't it?
And if not, shouldn't we ask engineering team to solve the BB issue?

joss-aztec:
> Isn't the simpler solution to have only deterministic opcodes?

That would be amazing if we could. Who's best to ask? Mike sounded frightened when I mentioned the requirement to fully parameterise pedersen.

> Isn't the simpler solution to have only deterministic opcodes?
> I believe we are close to ultra-plonk integration, which should give us that, shouldn't it?

This work decouples ACVM from baking in a specific proving system implementation (see #134 for motive). Directly integrating ultra-plonk takes us in the opposite direction (unless we're able to only compile in just its hypothetical implementation of the standardised deterministic opcode solvers without the rest of the system).

guipublic:
> unless we're able to only compile in just its hypothetical implementation of the standardised deterministic opcode solvers without the rest of the system

That's my understanding. I may be wrong but I believe the only problematic opcode is Pedersen.
UP Pedersen will be different than TP Perdersen but according to Zac, should be the version we want to support officially. The best person to ask should be Kev, but we should be able to implement UP Pedersen in Rust more easily than TP Pedersen and I think it worths the effort.

guipublic:
By the way, is there a reason why ACVM does not call the aztec backend for solving pedersen gate?
I knew (but don't remember the details)  it was an issue with wasm/native bb, but can we not solve this issue now?

joss-aztec:
This stepwise pattern is being abandoned in favour of the one used in https://github.com/noir-lang/acvm/pull/149 for resolving oracles. Furthermore, the work for decoupling ACVM from the backend has been deprioritised.

PR #137, opened 2023-03-07
feat!: Log Directive support for traces

### LINKED ISSUE #95: Switch PWG to return a list of logs

vezenovm:
# Related issue(s)

Resolves #95 

# Description

`Directive::Log` no longer prints the evaluated log while it is being evaluating during partial witness generation. This enables us to have both trace functionality and move logic for displaying logs to the code that calls the PWG rather than keeping it within the solver itself.

## Summary of changes

solve under the PWG now takes in a vector of `SolvedLog` structs. These solved logs contain either be a finalized string (calculated during compilation) or a vector of field elements that should be handled by the caller of `solve`. 

Several more changes showing how to handle logs correctly will be supplied in nargo (where we call `solve`)

(If applicable.)

## Test additions / changes

As this reduces the functionality of the log directive, implementation of tests can be seen in the noir/nargo PR that implements these changes.

# Checklist

# Additional context

(If applicable.)


acir/src/circuit/directives.rs
```diff
@@ -239,15 +247,26 @@ impl Directive {
...
-// evaluation, we can form an output string during ACIR generation.
-// Otherwise, we must store witnesses whose values will
-// be fetched during the PWG stage.
-pub enum LogInfo {
+/// This info is used when solving the initial witness
+/// If values are compile time and/or known during
+/// evaluation, we can form an output string during ACIR generation.
+/// Otherwise, we must store witnesses whose values will
+/// be fetched during the PWG stage.
+pub enum LogOutputInfo {
```
guipublic, line 255: I don't understand why it is one or the other, why can't you have some of the witness solved during ACIR generation and some of them not solved?

acir/src/circuit/directives.rs
```diff
@@ -239,15 +247,26 @@ impl Directive {
...
 }
 
+pub struct SolvedLog {
+    pub is_trace: bool,
+    pub output_info: SolvedLogOutputInfo,
+}
+
+pub enum SolvedLogOutputInfo {
+    FinalizedOutput(String),
+    WitnessValues(Vec<FieldElement>),
```
guipublic, line 267: If the log is solved, why do you keep the values and not just have a string?

acvm/src/lib.rs
```diff
@@ -74,7 +79,11 @@ pub trait PartialWitnessGenerator {
                         Self::solve_black_box_function_call(initial_witness, bb_func)
                     }
                     Opcode::Directive(directive) => {
-                        Self::solve_directives(initial_witness, directive)
+                        Self::solve_directives(initial_witness, directive).map(|possible_log| {
```
guipublic, line 82: I don't think the solve should take care of the logs, this should be handled internally by the log directive.

guipublic (changes requested):
I think ACIR generation should produce only one log directive, so that we do not try to solve multiple times the same witness.
Then this solve directive would have a vector of 'field or witness', depending on whether the requested log was solved during acir generation.
PWG can completely ignore this directive, because the caller knowing the solved witness and the directive can figure out on its own.
This keeps ACVM simpler.

acir/src/circuit/directives.rs
vezenovm: There is certain situations where ACIR generation does not solve the witness. For example, when implementing the log directive for the first time it was discovered that attempting to print the output of blackbox functions required waiting until PWG to fetch the witness value.

acir/src/circuit/directives.rs
vezenovm: I didn't want to leave any decoding of the field elements. Transforming `Vec<FieldElement>` into a string right away would also require the extra logic that currently exists in the ACVM where we check whether we have a singular FieldElement or [multiple](https://github.com/noir-lang/acvm/blob/0b305b7bc18cc9f1569b4ddb09ac50a48ecef5f9/acvm/src/pwg/directives.rs#L172) (as the string looks different based on whether we have a singular field element or an array).

acvm/src/lib.rs
vezenovm: I can move this logic to be internal to `solve_directives`

vezenovm:
> I think ACIR generation should produce only one log directive, so that we do not try to solve multiple times the same witness.
> Then this solve directive would have a vector of 'field or witness', depending on whether the requested log was solved during acir generation.
> PWG can completely ignore this directive, because the caller knowing the solved witness and the directive can figure out on its own.
> This keeps ACVM simpler.

I see what you mean about how the caller knowing the solved witness and the directive can figure out what to call on its own.  This is something I hadn't considered and will look at making this change here and the upstream Noir PR. 

I have a couple clarifying questions:

When you refer to solve directive do you mean `Directive::Log` itself or the new `SolvedLog` type I made in this PR? `SolvedLog` isn't necessarily another directive just a new type that we return for solved logs. 

If we were to only have a vector of fields and witnesses, the log directive would then require some extra information such as whether the vector of fields/witnesses is a string or a normal array of fields. You can see how I currently handle this in the Noir PR when evaluating [array logs](https://github.com/noir-lang/noir/pull/961/files#diff-1c46fe8b1afaf36ca27510239b464f4a4b4969ce805e034a5d61b791c040e079R323). I foresee if I made this change that `Directive::Log` would have to have another field similar to `trace_label` called `is_string_output`. I think this is ok though as we already have to store extra information in order to implement traces and the caller would be the one handling any log directive flags. What do you think?

vezenovm:
The VM has diverged a lot since this was opened, and these changes are small. Closing in favor of a fresh PR.

PR #138, opened 2023-03-10
chore(acir): public block id 

guipublic:
# Description

## Summary of changes

Make the id of BlockId public so we can create BlockId from a number.

# Checklist

# Additional context

We need to create the blocks in noir from the arrays

### DIFF
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -7,7 +7,7 @@ use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
-pub struct BlockId(u32);
+pub struct BlockId(pub u32);
 
 /// Operation on a block
 /// We can either write or read at a block index


jfecher (commented):
What functionality is this needed for?

acir/src/circuit/opcodes.rs
```diff
@@ -7,7 +7,7 @@ use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
-pub struct BlockId(u32);
+pub struct BlockId(pub u32);
```
phated, line 10: Would this be implemented with a `from<u32>` so the way it is stored isn't leaked to consumers/creators?

guipublic:
> What functionality is this needed for?

The opcode is created by noir, so it needs to set the id. This is currently impossible as the constructor is private.

acir/src/circuit/opcodes.rs
guipublic: I don't see the problem to leak this.

guipublic: No it won't change, I would not put anything else inside an id.

jfecher (approved):
Looks fine to me. We should usually be moving away from exposing internals of IDs in favor of keeping them opaque but it is indeed sometimes necessary to expose them. It is not worth keeping open such a simple PR, lets merge it.

PR #146, opened 2023-03-20
feat!(acir): Add RAM and ROM opcodes

guipublic:
# Description

## Summary of changes

Add ROM and RAM opcodes in order to support dynamic arrays for Aztec Backend.
The BLOCK opcode handles conditional writes by using an expression for the memory operation, but Barretenberg does not support this. We add RAM and ROM opcodes to help the backend to specify what it does support, so that noir can decide which opcode to use based on the array usage and the supported opcodes:

ROM: an initialisation phase with writes on constant indexes, then only read access
RAM: an initialisation phase with writes on constant indexes, then dynamic read or write access
BLOCK: the most flexible, if the backend supports it, it must enforce the correct constraints.

If the backend supports BLOCK, noir will use this opcode because it is more efficient,
if the backend supports RAM/ROM, then noir will use RAM/ROM opcode and generate conditional writes accordingly,
if the backend does not support any memory opcode, then noir will use the BLOCK opcode along with the memory consistency constraints.

# Checklist

### DIFF
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -7,7 +7,7 @@ use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
-pub struct BlockId(u32);
+pub struct BlockId(pub u32);
 
 /// Operation on a block
 /// We can either write or read at a block index
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -6,16 +6,55 @@ use crate::serialization::{read_n, read_u16, read_u32, write_bytes, write_u16, w
 use crate::BlackBoxFunc;
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Copy, Default, Debug)]
 pub struct BlockId(pub u32);
 
 /// Operation on a block
 /// We can either write or read at a block index
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
 pub struct MemOp {
     pub operation: Expression,
     pub index: Expression,
     pub value: Expression,
+}
+
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MemoryBlock {
+    pub id: BlockId,
+    pub len: u32,          //length of the memory block
+    pub trace: Vec<MemOp>, //trace of memory operations
+    pub init: u32,         //part of the trace dedicated to initialisation of the block
+}
+
+impl MemoryBlock {
+    pub fn read<R: Read>(mut reader: R) -> std::io::Result<Self> {
+        let id = read_u32(&mut reader)?;
+        let len = read_u32(&mut reader)?;
+        let init = read_u32(&mut reader)?;
+        let trace_len = read_u32(&mut reader)?;
+        let mut trace = Vec::with_capacity(len as usize);
+        for _i in 0..trace_len {
+            let operation = Expression::read(&mut reader)?;
+            let index = Expression::read(&mut reader)?;
+            let value = Expression::read(&mut reader)?;
+            trace.push(MemOp { operation, index, value });
+        }
+        Ok(MemoryBlock { id: BlockId(id), len, trace, init })
+    }
+
+    pub fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
+        write_u32(&mut writer, self.id.0)?;
+        write_u32(&mut writer, self.len)?;
+        write_u32(&mut writer, self.init)?;
+        write_u32(&mut writer, self.trace.len() as u32)?;
+
+        for op in &self.trace {
+            op.operation.write(&mut writer)?;
+            op.index.write(&mut writer)?;
+            op.value.write(&mut writer)?;
+        }
+        Ok(())
+    }
 }
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -24,7 +63,9 @@ pub enum Opcode {
     BlackBoxFuncCall(BlackBoxFuncCall),
     Directive(Directive),
     // Abstract read/write operations on a block of data
-    Block(BlockId, Vec<MemOp>),
+    Block(MemoryBlock),
+    ROM(MemoryBlock),
+    RAM(MemoryBlock),
 }
 
 impl Opcode {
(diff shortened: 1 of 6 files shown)


acir/src/circuit/opcodes.rs
```diff
@@ -4,27 +4,91 @@ use super::directives::{Directive, LogInfo};
...
+}
+
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Opcode {
     Arithmetic(Expression),
     BlackBoxFuncCall(BlackBoxFuncCall),
     Directive(Directive),
     // Abstract read/write operations on a block of data
-    Block(BlockId, Vec<MemOp>),
+    Block(MemoryBlock),
```
kevaundray, line 89: can you add an explanation as to what Block, RAM and ROM are and also why ACIR needs three of these memory handling opcodes?

acir/src/circuit/opcodes.rs
```diff
@@ -4,27 +4,91 @@ use super::directives::{Directive, LogInfo};
...
+
+        for op in &self.trace {
+            op.operation.write(&mut writer)?;
+            op.index.write(&mut writer)?;
+            op.value.write(&mut writer)?;
+        }
+        Ok(())
+    }
+
+    /// Returns the initialisation vector of the MemoryBlock
```
kevaundray, line 57: note: spellcheck

acir/src/circuit/opcodes.rs
```diff
@@ -4,27 +4,91 @@ use super::directives::{Directive, LogInfo};
...
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
 pub struct MemOp {
     pub operation: Expression,
     pub index: Expression,
     pub value: Expression,
 }
 
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MemoryBlock {
```
kevaundray, line 23: Can you add a comment to say what this?

acir/src/circuit/opcodes.rs
```diff
@@ -4,27 +4,91 @@ use super::directives::{Directive, LogInfo};
...
+    pub id: BlockId,
+    pub len: u32,          //length of the memory block
```
kevaundray, line 25: Note: for comments, lets put them ontop of the field with three slashes.

From:

```rust
pub len : u32, // length of the memory block 
```

to:

```rust
/// Length of the memory block
pub len : u32 
```

The latter one also shows up in your IDE when you hover over the variable in another file

acvm/src/compiler/transformers/fallback.rs
```diff
@@ -6,47 +6,50 @@ use acir::{
...
         let mut witness_idx = acir.current_witness_index + 1;
 
         for opcode in acir.opcodes {
-            let bb_func_call = match &opcode {
-                Opcode::Arithmetic(_) | Opcode::Directive(_) | Opcode::Block(_, _) => {
-                    // directive, arithmetic expression or block are handled by acvm
+            match &opcode {
+                Opcode::Arithmetic(_)
+                | Opcode::Directive(_)
+                | Opcode::Block(_)
```
kevaundray, line 27: Whats the difference between a directive and Block/RAM/ROM?

acvm/src/compiler/transformers/fallback.rs
```diff
@@ -6,47 +6,50 @@ use acir::{
...
                 Opcode::BlackBoxFuncCall(bb_func_call) => {
                     // We know it is an black box function. Now check if it is
                     // supported by the backend. If it is supported, then we can simply
                     // collect the opcode
-                    if is_supported(&bb_func_call.name) {
+                    if is_supported(&opcode) {
                         acir_supported_opcodes.push(opcode);
                         continue;
+                    } else {
+                        // If we get here then we know that this black box function is not supported
```
kevaundray, line 42: since we changed blackbox function above -- we should change mention of it here to

acvm/src/lib.rs
```diff
@@ -233,21 +235,25 @@ pub fn hash_constraint_system(cs: &Circuit) -> [u8; 32] {
 // This is set to match the previous functionality that we had
 // Where we could deduce what black box functions were supported
```
kevaundray, line 236: should change mention of black box function, since we changed the name

acvm/src/lib.rs
```diff
@@ -233,21 +235,25 @@ pub fn hash_constraint_system(cs: &Circuit) -> [u8; 32] {
...
 // by knowing the np complete language
-pub fn default_is_black_box_supported(
+pub fn default_is_opcode_supported(
     language: Language,
-) -> compiler::transformers::IsBlackBoxSupported {
+) -> compiler::transformers::IsOpcodeSupported {
     // R1CS does not support any of the black box functions by default.
```
kevaundray, line 241: same here

acvm/src/lib.rs
```diff
@@ -233,21 +235,25 @@ pub fn hash_constraint_system(cs: &Circuit) -> [u8; 32] {
...
     // The compiler will replace those that it can -- ie range, xor, and
-    fn r1cs_is_supported(_opcode: &BlackBoxFunc) -> bool {
+    fn r1cs_is_supported(_opcode: &Opcode) -> bool {
         false
     }
 
     // PLONK supports most of the black box functions by default
```
kevaundray, line 247: same here re black box function being changed to opcode

kevaundray (commented):
Left some general comments -- I wonder if we can reduce complexity by not trying to support Block with RAM and ROM -- ie maybe its possible for Ultraplonk to see a Block opcode and transform it into a structure that it can consume

guipublic:
> Left some general comments -- I wonder if we can reduce complexity by not trying to support Block with RAM and ROM -- ie maybe its possible for Ultraplonk to see a Block opcode and transform it into a structure that it can consume

Yes I thought about this but in the context of the backend. For the backend it would mean creating constraints which is not the right place, so it is better for noir to handle this. In order for noir to handle this, it needs to know what the backend support, hence the need for the RAM/ROM opcodes

Now, Ultra-Plonk could handle the BLOCK opcode as it can create constraints, but this mechanism can be useful for other backends so it makes sense for me to have in noir.
Furthermore UP cannot not support the full flexibility of the opcode as it requires the initialisation phase. This means that in theory we can get an error in BB which is not nice (in practise we always have the initialisation phase for now).

acvm/src/compiler/transformers/fallback.rs
guipublic: If the opcode is supported by the backend, the backend will layout the constraints for the memory block
If the opcode is not supported by the backend, noir will layout the constraints
In both cases, the solver needs to resolve the witness used in the memory trace.

acvm/src/compiler/transformers/fallback.rs
guipublic: I think it is ok, we are in the case of a black box function.

acir/src/circuit/opcodes.rs
guipublic: Done

acir/src/circuit/opcodes.rs
guipublic: Done

acir/src/circuit/opcodes.rs
guipublic: Done

acvm/src/lib.rs

acvm/src/lib.rs

acvm/src/lib.rs

PR #190, opened 2023-04-10
chore!: organise operator implementations for `Expression`

TomAFrench:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

The disorganised approach we have to implementing various traits on `Expression` has been bugging me. We've got a jumble of `From`s, maths operators and sorting operators all in the same file. This means that we're missing a few operator definitions which we'd want but it's not obvious that they're missing because there's not a defined order for them.

This PR takes `arithmetic.rs` and renames it to `expression.rs`, we then break this into three files:
- `mod.rs` which contains the bulk of the definition of `Expression` along with the main `impl`.
- `operators.rs` which defines `Add, Mul, Neg, Sub` on `Expression` against itself, `Witness` and `FieldElement`.
- `ordering.rs` which contains the logic necessary to implement `Ord/PartialOrd`

I've also made the `Add, Mul, Sub` operators all commutative.

One thing to note is that I've removed the from/operator implementations for `&FieldElement` and `&Witness` in favour of `FieldElement` and `Witness`. This is to match the fact that we only define arithmetic operators for `FieldElements` and not  `&FieldElement` in `acir_field` and `Witness` is a glorified `u32` so the benefits of passing by reference is going to be negligible or negative.

Breaking change as we'll likely have to remove some `&` to account for this.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/native_types/expression/mod.rs
+++ b/acir/src/native_types/expression/mod.rs
@@ -2,9 +2,10 @@ use crate::native_types::Witness;
 use crate::serialization::{read_field_element, read_u32, write_bytes, write_u32};
 use acir_field::FieldElement;
 use serde::{Deserialize, Serialize};
-use std::cmp::Ordering;
 use std::io::{Read, Write};
-use std::ops::{Add, Mul, Neg, Sub};
+
+mod operators;
+mod ordering;
 
 // In the addition polynomial
 // We can have arbitrary fan-in/out, so we need more than wL,wR and wO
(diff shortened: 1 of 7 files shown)


kevaundray:
I think some or all of the methods in ordering.rs won't be needed in the future, but for now, since we have them this PR generally makes sense

kevaundray (approved):

PR #194, opened 2023-04-12
feat!: Return entire unresolved Brillig opcode during oracle pause

vezenovm:
# Related issue(s)

Resolves (link to issue)

# Description

The ACVM pauses due to needing oracle outputs. When this was a singular opcode it was fine to simply return this oracle and then the ACVM caller can easily construct the new list of opcodes to be resolved. With Brillig we need the inputs and outputs of the brillig VM itself in order to process the opcode correctly. This PR adds the Brillig opcode as an unresolved opcode when we hit a Brillig oracle pause. This lets the ACVM caller to alter the brillig opcode without knowing or interacting with the encapsulating Brillig opcode's inputs and outputs. 

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -138,6 +138,7 @@ pub trait PartialWitnessGenerator {
                         stalled = false;
                         // InProgressBrillig Oracles must be externally re-solved
                         unresolved_brillig_oracles.push(oracle_wait_info);
+                        unresolved_opcodes.push(opcode.clone());
                     }
                     Ok(OpcodeResolution::Stalled(not_solvable)) => {
                         if opcode_not_solvable.is_none() {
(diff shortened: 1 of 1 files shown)


joss-aztec:
I see your test assumes that the brillig opcode will be at the front of `unresolved_opcodes`, but that's not actually a guarantee right? E.g. `solve` might have first skipped some incomplete arithmetic expression. I think it would be better if the brillig oracle and brillig parent opcode were surfaced as a pair. (This would also help to disambiguate the relation supposing that the inner for loop happened to hit two separate brillig bytecodes containing oracles in the same sweep of `opcode_to_solve`.)

PR #201, opened 2023-04-17
feat!: Use OracleOutput enum in OracleData

vezenovm:
# Related issue(s)

Resolves (link to issue)

# Description

This PR switches oracle output from a single register index to an enum. With only one register index we assume that if an oracle returns multiple complex types these are all flattened into one array and we are currently handling oracle outputs always as an array. This also means we break arrays of size 1 as there is currently an assumption that if there is only one output value supplied to the oracle it must be a single register. We now require someone to differentiate between register and array oracle outptus and have a `Vec<OracleOutput>` as part of OracleData.

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -359,7 +359,7 @@ mod test {
     use acir::{
         brillig_bytecode,
         brillig_bytecode::{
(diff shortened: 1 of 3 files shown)


jfecher (approved):
👍

guipublic (approved):

PR #207, opened 2023-04-18
feat: implement `add_mul` on `Expression`

TomAFrench:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

This PR migrates the remaining `Expression` arithmetic from [noirc_evaluator](https://github.com/noir-lang/noir/blob/809b85f751bd0e27ce8c4b38354bc051471d8522/crates/noirc_evaluator/src/ssa/acir_gen/constraints.rs#L55-L230). The aim behind this is so that we can remove some [very verbose usage of `add` and `subtract`](https://github.com/search?q=repo%3Anoir-lang%2Fnoir+subtract%28+add%28+language%3ARust&type=code&l=Rust) 

The `add` function has been changed to be the `add_mul` method on `Expression` so `a + k * b` can be called as `a.add_mul(k, b)`. I'm toying with the idea that this would be better named as `mul_add` as we're first multiplying then adding so this may be clearer. `add_mul` has had some refactoring applied to improve readability.

The `+` and `-` operator implementations have been updated to use `add_mul` under the hood as we should always have `a + b == a.add_mul(1, b)` which may not be the case otherwise.

I've also implemented `*` which replaces the `mul` function from noirc_evaluator.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/native_types/expression/mod.rs
+++ b/acir/src/native_types/expression/mod.rs
@@ -2,6 +2,7 @@ use crate::native_types::Witness;
 use crate::serialization::{read_field_element, read_u32, write_bytes, write_u32};
 use acir_field::FieldElement;
 use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
 use std::io::{Read, Write};
 
 mod operators;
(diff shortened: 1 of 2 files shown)


kevaundray (approved):
I'm happy with moving these into acvm, I think we need to have a clean-up of the logic inside of these functions, in particular the fact that we need `Ordering` seems like an abstraction leak. This is not a blocker for this PR though

PR #223, opened 2023-04-21
feat!: require `Backend` to implement `Default` trait

TomAFrench:
# Related issue(s)

Resolves (link to issue)

# Description

In our new merged `acvm-backend-barretenberg` crate, we're implementing the `Backend` trait directly on the `Barretenberg` struct rather than having a separate `Plonk` unit struct which then constructs `Barretenberg` when needed.

https://github.com/noir-lang/aztec_backend/blob/9fed6e52778d993d7742c46bbeec66087958c075/acvm_backend_barretenberg/src/lib.rs#L21-L26

This means that when we're actually using the backend in Nargo we need to construct the struct, i.e we need to call `let backend = Barretenberg::default()` rather than just being able to do `let backend = Plonk`.

Example: https://github.com/noir-lang/noir/commit/ad822e661ce4082ed96f790e8baa2f76d460b9fc

We then want to enforce that this method exists on all backends so it must be included in the trait definition.

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -92,7 +92,7 @@ fn first_missing_assignment(
     })
 }
 
-pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator {}
+pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default {}
 
 /// This component will generate the backend specific output for
 /// each OPCODE.
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -318,8 +318,8 @@ mod test {
     };
 
     use crate::{
-        pwg::block::Blocks, Backend, OpcodeResolution, OpcodeResolutionError,
-        PartialWitnessGenerator, PartialWitnessGeneratorStatus,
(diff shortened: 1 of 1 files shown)


acvm/src/lib.rs
```diff
@@ -92,7 +92,7 @@ fn first_missing_assignment(
     })
 }
 
-pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator {}
+pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default {}
```
phated, line 95: Does default allow us to do some setup? I think some backends probably need to do setup work in the function so maybe this should be a `new() -> Self` function if default doesn't allow setup work

TomAFrench: What sort of setup would you want to do that can't be done in `default()`? afaik they're directly equivalent.

My choice for `Default` over `new() -> Self` is that:
a. it's a standardized trait whereas `new()` is an ad hoc method.
b. we can let backends define their own `new` function which takes additional methods for e.g. injecting a CRS or something.

PR #225, opened 2023-04-23
fix: prevent `bn254` feature flag always being enabled

TomAFrench:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

Currently we're always turning on the `bn254` field due to default feature flags, this makes it impossible to use any different fields. I've fixed the feature flag setup so that only the field supplied to the `acvm` package gets turned on.

The second commit contains some miscellaneous cleanup of the various toml files.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir_field/Cargo.toml
+++ b/acir_field/Cargo.toml
@@ -26,6 +26,5 @@ ark-ff = { version = "^0.4.0", optional = true, default-features = false }
 cfg-if = "1.0.0"
 
 [features]
-default = ["bn254"]
 bn254 = ["ark-bn254", "ark-ff"]
 bls12_381 = ["ark-bls12-381", "ark-ff"]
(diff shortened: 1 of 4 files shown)


acir_field/Cargo.toml
```diff
@@ -27,5 +27,5 @@ cfg-if = "1.0.0"
 
 [features]
 default = ["bn254"]
```
kevaundray, line 29: why not just remove the default from here?

acvm/Cargo.toml
```diff
@@ -30,8 +30,8 @@ indexmap = "1.7.0"
 thiserror = "1.0.21"
 
 [features]
-bn254 = ["acir/bn254"]
-bls12_381 = ["acir/bls12_381"]
+bn254 = ["acir/bn254", "stdlib/bn254"]
```
kevaundray, line 33: Passing comment: similar to how we re-export acir from acvm and acir_field, we may want to eventually re-export stdlib too

kevaundray (commented):
Looks good to me -- left some comments

kevaundray (approved):

PR #231, opened 2023-04-26
feat(acvm)!: Add CommonReferenceString backend trait

phated:
# Related issue(s)

Needed for https://github.com/noir-lang/acvm-backend-barretenberg/issues/160

# Description

## Summary of changes

This requires a `CommonReferenceString` trait for ACVM Backends.

The reason some functions in other traits were modified to take the `common_reference_string` is because backends can have a circular dependency, such that they need to instantiate themselves before they can get the circuit size for a given circuit and the CRS often needs the circuit size to the data it is using. This means we can't do `Backend::new(CRS::new(circuit_size))` and resulted the functions that need a CRS taking it as an argument.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -85,8 +85,8 @@ pub trait CommonReferenceString {
     ///
     /// This function will be called if the common reference string has been cached previously
     /// and the backend can update it if necessary. This may happen if the common reference string
-    /// contains fewer than the number of points needed by the circuit, or any other checks the backend
-    /// must do.
+    /// contains fewer than the number of points needed by the circuit, or fails any other checks the backend
+    /// must perform.
     ///
     /// If the common reference string doesn't need any updates, implementors can return the value passed.
     async fn update_common_reference_string(
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -65,6 +65,11 @@ pub trait Backend:
 {
 }
 
+// Unfortunately, Rust doesn't natively allow async functions in traits yet.
+// So we need to annotate our trait with this macro and backends need to attach the macro to their `impl`.
+//
+// For more details, see https://docs.rs/async-trait/latest/async_trait/
+// and https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/
 #[async_trait]
 pub trait CommonReferenceString {
     /// The Error type returned by failed function calls in the CommonReferenceString trait.
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -19,6 +19,9 @@ use core::fmt::Debug;
 use std::collections::BTreeMap;
 use thiserror::Error;
 
+// We re-export async-trait so consumers can attach it to their impl
+pub use async_trait::async_trait;
+
 // re-export acir
 pub use acir;
 pub use acir::FieldElement;
@@ -54,8 +57,39 @@ pub enum OpcodeResolutionError {
 }
 
 pub trait Backend:
-    SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default + Debug
+    SmartContract
+    + ProofSystemCompiler
+    + PartialWitnessGenerator
+    + CommonReferenceString
+    + Default
+    + Debug
 {
+}
+
+#[async_trait]
+pub trait CommonReferenceString {
+    /// The Error type returned by failed function calls in the CommonReferenceString trait.
+    type Error: std::error::Error; // fully-qualified named because thiserror is `use`d at the top of the crate
+
+    /// Provides the common reference string that is needed by other traits
+    async fn generate_common_reference_string(
+        &self,
+        circuit: &Circuit,
+    ) -> Result<Vec<u8>, Self::Error>;
+
+    /// Updates a cached common reference string within the context of a circuit
+    ///
+    /// This function will be called if the common reference string has been cached previously
+    /// and the backend can update it if necessary. This may happen if the common reference string
+    /// contains fewer than the number of points needed by the circuit, or any other checks the backend
+    /// must do.
+    ///
+    /// If the common reference string doesn't need any updates, implementors can return the value passed.
+    async fn update_common_reference_string(
+        &self,
+        common_reference_string: Vec<u8>,
+        circuit: &Circuit,
+    ) -> Result<Vec<u8>, Self::Error>;
 }
 
 /// This component will generate the backend specific output for
@@ -157,8 +191,12 @@ pub trait SmartContract {
 
     // TODO: Allow a backend to support multiple smart contract platforms
 
-    /// Returns an Ethereum smart contract to verify proofs against a given verification key.
-    fn eth_contract_from_vk(&self, verification_key: &[u8]) -> Result<String, Self::Error>;
+    /// Returns an Ethereum smart contract to verify proofs against a given common reference string and verification key.
+    fn eth_contract_from_vk(
+        &self,
+        common_reference_string: &[u8],
+        verification_key: &[u8],
+    ) -> Result<String, Self::Error>;
 }
 
 pub trait ProofSystemCompiler {
@@ -179,13 +217,18 @@ pub trait ProofSystemCompiler {
 
     /// Generates a proving and verification key given the circuit description
     /// These keys can then be used to construct a proof and for its verification
-    fn preprocess(&self, circuit: &Circuit) -> Result<(Vec<u8>, Vec<u8>), Self::Error>;
+    fn preprocess(
+        &self,
+        common_reference_string: &[u8],
+        circuit: &Circuit,
+    ) -> Result<(Vec<u8>, Vec<u8>), Self::Error>;
 
     /// Creates a Proof given the circuit description, the initial witness values, and the proving key
     /// It is important to note that the intermediate witnesses for black box functions will not generated
     /// This is the responsibility of the proof system.
     fn prove_with_pk(
         &self,
+        common_reference_string: &[u8],
         circuit: &Circuit,
         witness_values: BTreeMap<Witness, FieldElement>,
         proving_key: &[u8],
@@ -194,6 +237,7 @@ pub trait ProofSystemCompiler {
     /// Verifies a Proof, given the circuit description, the circuit's public inputs, and the verification key
     fn verify_with_vk(
         &self,
+        common_reference_string: &[u8],
         proof: &[u8],
         public_inputs: BTreeMap<Witness, FieldElement>,
         circuit: &Circuit,
--- a/acvm/Cargo.toml
+++ b/acvm/Cargo.toml
@@ -29,6 +29,7 @@ k256 = { version = "0.7.2", features = [
 ] }
 indexmap = "1.7.0"
 thiserror = "1.0.21"
+async-trait = "0.1"
 
 [features]
 default = ["bn254"]


phated:
Opened as a draft to discuss

acvm/src/lib.rs
```diff
@@ -92,7 +92,18 @@ fn first_missing_assignment(
...
+    SmartContract + ProofSystemCompiler + PartialWitnessGenerator + ReferenceString + Default
+{
+}
+
+pub trait ReferenceString {
+    type ReferenceString; // TODO: Figure out serde trait
+
+    // TODO: Figure out Err type
+    // TODO: Figure out async trait
+    fn get_reference_string(&self, circuit: &Circuit) -> Result<Self::ReferenceString, String>;
```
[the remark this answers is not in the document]
phated: proposal: `can_compute_circuit(&self, reference_string: &Self::ReferenceString, circuit: &Circuit)`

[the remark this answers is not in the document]
phated: Though, the above proposed API doesn't account for appending more `g1_data` to the data we already have; instead opting to download all of the data to fulfill a larger circuit.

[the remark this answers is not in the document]
vezenovm: It sounds like we could adapt `get_reference_string` or include a new method to account for appending data rather than re-downloading data we already have?

acvm/src/lib.rs
```diff
@@ -92,7 +97,26 @@ fn first_missing_assignment(
...
 }
 
-pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default {}
+pub trait Backend:
+    SmartContract + ProofSystemCompiler + PartialWitnessGenerator + ReferenceString + Default
+{
+}
+
+#[async_trait]
+pub trait ReferenceString {
```
guipublic, line 106: Why do we need a new trait for this?
I think all backends should implement this interface. If a backend is not using the crs model, it can easily returns an empty vector.

acvm/src/lib.rs
```diff
@@ -92,7 +97,26 @@ fn first_missing_assignment(
...
+    /// The Error type returned by failed function calls in the ReferenceString trait.
+    type Error: std::error::Error; // fully-qualified named because thiserror is `use`d at the top of the crate
+
+    /// Provides the reference string that is needed by other traits
+    async fn get_reference_string(&self, circuit: &Circuit) -> Result<Vec<u8>, Self::Error>;
+
+    /// Validates a (likely cached) reference string against a circuit
+    fn is_reference_string_valid(
```
guipublic, line 114: what is this supposed to do?

acvm/src/lib.rs
```diff
@@ -92,7 +97,26 @@ fn first_missing_assignment(
...
+{
+}
+
+#[async_trait]
+pub trait ReferenceString {
+    /// The Error type returned by failed function calls in the ReferenceString trait.
+    type Error: std::error::Error; // fully-qualified named because thiserror is `use`d at the top of the crate
+
+    /// Provides the reference string that is needed by other traits
+    async fn get_reference_string(&self, circuit: &Circuit) -> Result<Vec<u8>, Self::Error>;
```
guipublic, line 111: I would use common_reference_string instead of just reference_string, and I would call the function generate_common_reference_string(). 
The backend can cache it if he wants to, or download it, .. these are implementation details that are not relevant for the backend interface.

acvm/src/lib.rs
```diff
@@ -222,13 +250,18 @@ pub trait ProofSystemCompiler {
 
     /// Generates a proving and verification key given the circuit description
     /// These keys can then be used to construct a proof and for its verification
-    fn preprocess(&self, circuit: &Circuit) -> Result<(Vec<u8>, Vec<u8>), Self::Error>;
+    fn preprocess(
```
guipublic, line 253: I think preprocess() can be confusing with preprocessing snarks. Normally it is the generator which computes the keys, so I would instead have these 2 functions:
generate_proving_key(crs, circuit)
generate_verification_key(crs, circuit)

Is there an advantage to do both at the same time?

guipublic (changes requested):

acvm/src/lib.rs
phated: Happy to rename it, but we're moving the caching into Nargo. Fetching or generating in some fashion is determine by the backend itself, but not filesystem access (helps with browser environment).

acvm/src/lib.rs
phated: I'm fine with that, but I'd split it into a different PR. Pinging @vezenovm because he added preprocess, I believe.

vezenovm: > Is there an advantage to do both at the same time?

To my knowledge proving schemes generally generate the verification key based upon the proving key (at least this is how it is done in barretenberg). Looking into the method @phated just linked, the method actually fetches the vk from the pk [internally](https://github.com/arkworks-rs/groth16/blob/9bac46fb3b1fbc0d2c88c6bc59b0ba798216bf6f/src/lib.rs#L71). Separating the preprocess method could lead to situations that require generating the proving key twice (which is the most intensive part of the preprocess operation) without making API changes to the proving system. Of course we don't want to base our API off of proving systems and rather those looking to integrate should change their API instead. However, I don't see the disadvantage of having preprocess be one method.

In our `aztec_backend` we have two separate composer methods to compute the proving key and the verification key (which takes in a proving key) so our API technically could support `generate_proving_key` and `generate_verification_key` without a performance degradation, but this does not look to be the case in all proving schemes. Having one `preprocess` method covers the case where a proving system supports generating the proving key and verification key separately, as well as at the same time. 

I would support adding two functions `generate_proving_key` and `generate_verification_key` as part of ACVM API, while keeping `preprocess`. The backend system could then chose if it wants to integrate these methods separately or just a single `preprocess` method. Either way I think this is entirely separate from the reference string work in this PR and should be a separate PR as it requires a new API change.

guipublic: I separated them because a verifier just need the verification key.
However, if in practise they are always computed at the same time, then I guess it is ok to have a generate() method, and may be some get_proving_key() and get_verification_key().

acvm/src/lib.rs
guipublic: How can a backend validate this if not generating it again and comparing with the input?

acvm/src/lib.rs
phated: Fetching or generating a CRS is done by a backend. Caching and filesystem access is done by Nargo.

acvm/src/lib.rs
guipublic: SmartContract makes sense, you do not need it to prove and verify.
Reference string is less clear, I guess you can generate a circuit for a backend, without having the CRS, but the ProofSystemCompiler has the preprocess interface, meaning it requires the ReferenceString.
I suggest to have a Generator trait instead which will deal with the CRS, proving key and verification key.

acvm/src/lib.rs
guipublic: The validity of the download should be done when it is downloaded, no need extra interface for this.
File corruption should be handled by the one caching it (in that case it is cargo), no need extra interface for this.
The circuit size test makes sense but it is dependent on the proving system so it is not generic. I am not sure if this works well for a groth16 backend?

acvm/src/lib.rs
vezenovm: Yeah I think we can discussing adding two `generate_proving_key` and `generate_verification_key` methods, but it is still a different scope than this PR

acvm/src/lib.rs
phated: But the SmartContract has a dependency on the CRS, so we should probably just throw out any separation and call it all a single Backend trait.

acvm/src/lib.rs
phated: A cached file can be corrupted by the filesystem, so you can't just do it once and forget about it.

acvm/src/lib.rs
phated: Renamed

acvm/src/lib.rs
phated: Since this API design is out of the scope of this PR, I opened an issue to continue the discussion: #256

acvm/src/lib.rs
guipublic: sure, but you can hash it and checks if it is corrupted or not.

acvm/src/lib.rs
guipublic: I don't see why you should implement SmartContract if you don't need it, so the separation here is good IMHO.
Generator and ProofSystem have different responsibilities, and Generator does not depend on ProofSystem so I am also fine with such separation.

phated: A backend requires SmartContract. You must implement all traits or you aren't a Backend.

acvm/src/lib.rs
phated: Who is hashing it? What are you comparing it to?

guipublic: you hash it when you retrieve it from the backend, and you compare this hash from the hash of the file

acvm/src/lib.rs
guipublic: I think this is bad, but it is another topic, and the trait separation cannot hurt, so I think it is a move in a good direction.

TomAFrench: In an ideal world, we'd make `SmartContract` optional as I imagine a lot of backends will leave this unimplemented. 
My concern with this is maintaining a consistent interface for nargo as currently we wouldn't be able to compile a backend which didn't implement the full `Backend` trait.

I suggest we park this for the pluggable backends push.

acvm/src/lib.rs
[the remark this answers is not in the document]
phated: I've changed the interface to be `generate_common_reference_string` and `update_common_reference_string` to account for this. It is worth noting that Nargo CLI will need to write data in every scenario with this API but it will lighten network loads, so I think it is a fine trade-off.

acvm/src/lib.rs
phated: #280

acvm/src/lib.rs
phated: This API has been changed to be `update_common_reference_string` to account for downloading only the additional CRS data that is needed (instead of the entirety of the data). It is worth noting that Nargo CLI will need to write data in every scenario with this API but it will lighten network loads, so I think it is a fine trade-off.

phated:
This is a significant change to Backends, so we should hold it (even after approvals) until the CRS changes all prepared for the Backend and Nargo.

acvm/src/lib.rs
```diff
@@ -54,10 +57,41 @@ pub enum OpcodeResolutionError {
...
+    async fn generate_common_reference_string(
+        &self,
+        circuit: &Circuit,
+    ) -> Result<Vec<u8>, Self::Error>;
+
+    /// Updates a cached common reference string within the context of a circuit
+    ///
+    /// This function will be called if the common reference string has been cached previously
+    /// and the backend can update it if necessary. This may happen if the common reference string
+    /// contains less than the number of points needed for by circuit, or any other checks the backend
```
TomAFrench, line 84:
```suggestion
    /// contains fewer than the number of points needed by the circuit, or any other checks the backend
```

nit

acvm/src/lib.rs
```diff
@@ -54,10 +57,41 @@ pub enum OpcodeResolutionError {
...
+    /// must do.
+    ///
+    /// If the common reference string doesn't need any updates, implementors can return the value passed.
+    async fn update_common_reference_string(
+        &self,
+        common_reference_string: &[u8],
+        circuit: &Circuit,
+    ) -> Result<Vec<u8>, Self::Error>;
```
TomAFrench, line 92:
```suggestion
    /// If the common reference string doesn't need any updates, implementors can return the value passed.
    async fn update_common_reference_string(
        &self,
        common_reference_string: Vec<u8>,
        circuit: &Circuit,
    ) -> Result<Vec<u8>, Self::Error>;
```

With the comment in mind, shouldn't we pass ownership over the CRS to the backend? This way is can just hand ownership right back whereas currently it needs to clone the vector.

TomAFrench, line 92:
```suggestion
    /// If the common reference string doesn't need any updates, implementors can return the value passed.
    async fn update_common_reference_string(
        &self,
        common_reference_string: Vec<u8>,
        circuit: &Circuit,
    ) -> Result<Vec<u8>, Self::Error>;
```

With the comment in mind, shouldn't we pass ownership over the CRS to the backend? This way it can just hand ownership right back whereas currently it needs to clone the vector.

TomAFrench (commented):
Looks good, couple of nits.

acvm/src/lib.rs
phated: I'm not sure handing it back matters because generally the backend is going to deserialize it to view it in their format and perform checks—such as the barretenberg backend turning it into the CRS struct to check the number of points.

I can see wanting to transfer the ownership from an API standpoint though, since a backend should never use that cached version again after calling this function 👍

vezenovm (approved):

guipublic (approved):

TomAFrench:
Will take a look at this in the morning (singapore time) and can merge after.

TomAFrench (approved):

TomAFrench:
Looks good, modified one comment.

PR #234, opened 2023-04-26
feat(acir)!: BlackBoxFunc::VerifyProof

### LINKED ISSUE #198: Add recursion opcode
### Problem

The ability to verify a program in another program is one feature which modern programming languages have. As with all opcodes, this can be done in the frontend using the arithmetic opcode, however it will be in-efficient and there are ways in which the proving system can do it in a more optimal way. One of these is delaying the pairing computation until verification, so it is not computed inside of the circuit.

### Proposed solution

**An alternative**

Lets first discuss an alternative which may eventually deprecate the strategy we will go with now.

The idea is to look at what parts of the verification algorithms we _really_ need to delegate to the proving system. In the case that the proving system uses [KZG](https://link.springer.com/chapter/10.1007/978-3-642-17373-8_11) we would only create a pairings opcode and the frontend will efficiently compute the rest. An example of a frontend would be Noir. It is the case that there are other components in the verification algorithm that are expensive like big integer arithmetic and this solution would hinge on us having enough opcodes such that they cover the possible usecases such as big integer.

The reason to not go with this strategy _right_ now is because we would need to do a survey of the known proving systems and figure out what components we really need to delegate to the proving system.

The advantages of this strategy is that you more of the code is implemented in the frontend or in the acvm-stdlib, making it less opaque.

**Proposed solution** 

To implement a `VERIFY_PROOF` opcode which will take a proof, public inputs and a verification key. It will then return true to indicate whether the proof is successful and false to indicate whether the proof is not successful.

**Structures**

This will be a new black box function, where the function definition will be:

```rust
FuncDefinition {
           name : "verify_proof",
           input_size: InputSize::Variable,
           output_size: OutputSize(1),
},
```

The `input_size` will be the concatenation of the verification key, proof and public inputs. The input size is variable for two reasons:

- The public inputs can vary depending on the circuit.
- (Not ideal) we don't specify what proving system we are verifying a proof for, so this recursion opcode can be used to verify plonk/groth16/bulletproofs for example. This is not ideal because if a frontend changes the backend but does not change the proof, then they will encounter a failing proof.

For the latter point, we could introduce an opcode for each proof system that we can recurse on though this is not an ideal solution because each time we add a backend which is supported, we would then need to add an opcode.

### Alternatives considered

_No response_

### Additional context

_No response_

### Submission Checklist

vezenovm:
# Related issue(s)

Resolves #198 

# Description

Add VerifyProof variant to BlackBoxFunc enum

## Summary of changes

This will be used by backends looking to integrate recursion. As per the issue discussion, this requires enabling the OutputSize of FuncDefinition's to be Variable. I left the `InputSize` and `OutputSize` enums distinct for now and left a comment under what scenarios OutputSize can be Variable. The OutputSize is NOT dependent upon the user inputs (like the input size to sha256 for example), but rather dependent upon the proving system that is integrating with ACVM. This distinction is important as this change to variable output size does imply support for dynamic circuits. 

I will make an issue to consider how we can best differentiate whether the variable OutputSize can be dependent upon a gadget's input. It might need to be in this PR though where we have a function in the ProofSystemCompiler similar to `black_box_function_supported` that can output a BlackBoxFunc's proof system dependent OutputSize. Opening this PR as part of the discussion. 

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/black_box_functions.rs
+++ b/acir/src/circuit/black_box_functions.rs
@@ -65,6 +65,7 @@ impl BlackBoxFunc {
             10 => BlackBoxFunc::XOR,
             11 => BlackBoxFunc::RANGE,
             12 => BlackBoxFunc::Keccak256,
+            13 => BlackBoxFunc::VerifyProof,
             _ => return None,
         };
         Some(function)
--- a/acir/src/circuit/black_box_functions.rs
+++ b/acir/src/circuit/black_box_functions.rs
@@ -22,6 +22,7 @@ pub enum BlackBoxFunc {
     EcdsaSecp256k1,
     FixedBaseScalarMul,
     Keccak256,
+    VerifyProof,
 }
 
 impl std::fmt::Display for BlackBoxFunc {
@@ -46,6 +47,7 @@ impl BlackBoxFunc {
             BlackBoxFunc::XOR => 10,
             BlackBoxFunc::RANGE => 11,
             BlackBoxFunc::Keccak256 => 12,
+            BlackBoxFunc::VerifyProof => 13,
         }
     }
     pub fn from_u16(index: u16) -> Option<Self> {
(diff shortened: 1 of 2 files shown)


vezenovm:
Converting to draft for now as this is not ready

vezenovm:
Closing in favor of #291. There were many breaking changes and it was simpler to cherry-pick against the new PR removing manual serialization that uses serde

PR #248, opened 2023-04-28
feat(acvm)!: Introduce Error type for fallible Backend traits

### LINKED ISSUE #235: ACVM Backend functions should be fallible
### Problem

Currently, a backend needs to panic if it does something fallible and that thing fails. This is not ideal because then Nargo or whatever other backend orchestrator is not in control of the errors.

### Proposed solution

Make all the functions return Result type. Define an error type that can be consumed by orchestrators.

### Alternatives considered

_No response_

### Additional context

_No response_

### Submission Checklist

phated:
# Related issue(s)

Resolves #235

# Description

## Summary of changes

This adds a required `Fallible` trait to ACVM backends. It is just used to create an Associated Type called Error that implements `std::error::Error` because traits aren't allowed to use `impl std::error::Error` and we don't want to require Backend implementors to have to Box all of their errors.

This uses the Error type to make other trait functions return Result types, which will allow Nargo to handle errors returned by a backend instead of the backend needing to panic. For example, the barretenberg wasm backend and https://github.com/TomAFrench/acvm-backend-groth16 both needed to unwrap/expect a lot of things but those errors should be bubbled up to nargo.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

Planning to implement this throughout aztec_backend and nargo.

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -92,15 +92,7 @@ fn first_missing_assignment(
     })
 }
 
-pub trait Backend:
-    SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default + Fallible
-{
-}
-
-pub trait Fallible {
-    /// The Error type returned by failed function calls in the Backend trait.
-    type Error: std::error::Error; // fully-qualified named because thiserror is `use`d at the top of the crate
-}
+pub trait Backend: SmartContract + ProofSystemCompiler + PartialWitnessGenerator + Default {}
 
 /// This component will generate the backend specific output for
 /// each OPCODE.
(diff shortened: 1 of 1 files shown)


vezenovm (approved):

TomAFrench:
I'm wondering whether we want to allow for multiple error types here? e.g. `ProofSystemCompiler` should return different errors to `SmartContract`?

We could also go more granular.

phated:
> I'm wondering whether we want to allow for multiple error types here? e.g. ProofSystemCompiler should return different errors to SmartContract?

I think it's fine to group them all under one error type. My prototype used a `thiserror` defined by the backend so it can provide whatever errors it needs.

phated:
As per @TomAFrench and @kevaundray, I've switched this to having a separate `type Error` on each trait that has fallible functions.

kevaundray (approved):

PR #251, opened 2023-05-01
feat(acvm): Add generic error for failing to solve an opcode

phated:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

This adds a generic `OpcodeSolveFailed` error to the `OpcodeNotSolvable` error enum so backends can report other problems with solving an opcode. I encountered this problem when converting aztec_backend to the fallible traits.

I think the data carried still needs to be bikeshed, since only an error string is probably not enough information.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -52,7 +52,7 @@ pub enum OpcodeResolutionError {
     UnexpectedOpcode(&'static str, BlackBoxFunc),
     #[error("expected {0} inputs for function {1}, but got {2}")]
     IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
-    #[error("failed to solve blackbox function: {0}, reason: {1}")] 
+    #[error("failed to solve blackbox function: {0}, reason: {1}")]
     BlackBoxFunctionFailed(BlackBoxFunc, String),
 }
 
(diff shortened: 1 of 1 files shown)


acvm/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub enum OpcodeResolutionError {
     UnexpectedOpcode(&'static str, BlackBoxFunc),
     #[error("expected {0} inputs for function {1}, but got {2}")]
     IncorrectNumFunctionArguments(usize, BlackBoxFunc, usize),
+    #[error("opcode solve failed: {0}")]
+    OpcodeSolveFailed(String),
```
kevaundray, line 56: This name is too generic, perhaps it could be changed to something like:

```rust
    #[error("failed to solve blackbox function: {0}, reason: {1}")]
     BlackBoxFunctionFailed(BlackBoxFunc, String),
```

The reason for this is because a generic error type may then become the place where all new errors get placed. We ideally want to know exactly what the context is around the error that was being thrown, so that if we decide to remove blackbox function solving for example, we can then remove the accompanying error types and backends will know immediately what code needs to be revised.

phated:
```suggestion
    #[error("failed to solve blackbox function: {0}, reason: {1}")] 
    BlackBoxFunctionFailed(BlackBoxFunc, String),
```
Yep, I'm fine with that; however, note that #257 will no longer pass the BlackBoxFunc to the backend but we can adjust the solve function to map the error to add the information

phated, line 56:
```suggestion
    #[error("failed to solve blackbox function: {0}, reason: {1}")] 
    BlackBoxFunctionFailed(BlackBoxFunc, String),
```

kevaundray (approved):

Issue #186, opened 2023-04-06
Consider changing the witness map serialisation method


TomAFrench:
### Problem

https://github.com/noir-lang/acvm/pull/177 removes the deprecated methods `Circuit::from_bytes` and `Circuit::to_bytes`.

https://github.com/noir-lang/acvm/blob/fd28657426260ce3c53517b75a27eb5c4a74e234/acir/src/circuit/mod.rs#L47-L66

This was chosen to give a serialisation format which is "easy to implement in many languages (without ffi)" however we still use the old serialisation method for witnesses.

https://github.com/noir-lang/acvm/blob/fd28657426260ce3c53517b75a27eb5c4a74e234/acir/src/native_types/witness.rs#L31-L48

This means that when manipulating witness maps in TS we receive this as opaque binary which requires us to use methods such as https://github.com/noir-lang/aztec_backend/pull/96 as opposed to performing this flattening in TS.

### Proposed solution

We should investigate alternative serialisation methods which would be easy to implement in other languages.

### Submission Checklist

- [X] Once I hit submit, I will assign this issue to the Project Board with the appropriate tags.

PR #252, opened 2023-05-01
feat!: Introduce WitnessMap data structure to avoid leaking internal structure

phated:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

This implements a `WitnessMap` type that wraps BTreeMap throughout the project. I noticed that this was a leaky abstraction because the `to_bytes` and `from_bytes` implementations on Witness took or returned BTreeMaps instead of operating on one Witness.

By implementing this newtype, we can implement traits against it to provide a much smoother API and still allow us to change the underlying data type if we want or need to. I also hope this streamlines serde between wasm & JS in the future but I've not built that out yet.

While I was updating Noir itself, I noticed that it also had a `WitnessMap` type alias, so that further solidified my belief that this type needed to exist in ACVM.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)


acir/src/native_types/witness.rs
```diff
@@ -27,23 +27,75 @@ impl Witness {
...
+
+impl From<u32> for Witness {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+/// A map from the witnesses in a constraint system to the field element values
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Default, Serialize, Deserialize)]
+pub struct WitnessMap(BTreeMap<Witness, acir_field::FieldElement>);
```
TomAFrench, line 40:
```suggestion
pub struct WitnessMap(BTreeMap<Witness, FieldElement>);
```

Can we use the non-qualified form here?

acir/src/native_types/witness.rs
```diff
@@ -27,23 +27,75 @@ impl Witness {
...
-        rmp_serde::from_slice(buf_d.as_slice()).unwrap()
+        Self(rmp_serde::from_slice(buf_d.as_slice()).unwrap())
+    }
+}
+
+impl From<WitnessMap> for Vec<acir_field::FieldElement> {
+    fn from(val: WitnessMap) -> Self {
+        val.0.into_values().collect()
     }
 }
```
TomAFrench, line 101: I'm a little cautious about implementing this as there isn't a single format for converting a `WitnessMap` into a `Vec<FieldElement>`.

e.g. when proving we fill in any missing indices with zeros: https://github.com/noir-lang/aztec_backend/blob/c9fb9e806f1400a2ff7594a0669bec56025220bb/src/acvm_interop/proof_system.rs#L78-L97 but when verifying we flatten the public inputs in the way you have here https://github.com/noir-lang/aztec_backend/blob/c9fb9e806f1400a2ff7594a0669bec56025220bb/src/acvm_interop/proof_system.rs#L66 

As there's not a single canonical conversion I don't think we can implement this trait.

acir/src/native_types/witness.rs
```diff
@@ -27,23 +27,75 @@ impl Witness {
...
+    }
+}
+
+impl From<BTreeMap<Witness, acir_field::FieldElement>> for WitnessMap {
+    fn from(value: BTreeMap<Witness, acir_field::FieldElement>) -> Self {
+        Self(value)
+    }
+}
+
+impl From<WitnessMap> for Vec<u8> {
```
TomAFrench, line 78: These should implement `TryFrom` rather than `From`

TomAFrench, line 78: I'm also unsure whether we want to keep this serialisation method in future as we've removed all the other instances of it: https://github.com/noir-lang/acvm/issues/186

TomAFrench (commented):
Nice! Yeah, this has been something floating in the back of my mind as well so thank you for implementing it.

Just a couple of comments on this. We should also check the `aztec_backend` and `noir` repos to ensure we're not missing out any features we need to add (I can't see any that aren't already covered here)

acir/src/native_types/witness.rs
phated: Was just following what the rest of the module did, but I can hoist it.

acir/src/native_types/witness.rs
phated: This was just the implementation needed for https://github.com/noir-lang/aztec_backend/pull/152/files#diff-38a8cdcfc939b53aa726b0f6a6a3c4940449409f27ffb409c36c64e4ea1b7749R65 since I didn't want to leak the `into_keys` or `into_values` functions which seem a little too specific to the internal data structure.

acir/src/native_types/witness.rs
phated: Yes, I built this on top of 0.10.3 instead of the pre-0.11 so we don't have fallible functions everywhere so I figured unwrap is easiest. Probably need to finish and land https://github.com/noir-lang/aztec_backend/pull/151 so I can avoid unwrapping.

phated: Will that mean that `read` and `write` need to be implemented on the WitnessMap type? I believe that is what is used for Circuits now

TomAFrench: Nah, I think at most we'd keep this as an ad hoc method so we can add a deprecation notice in future.

phated: Hm 🤔  I'm currently thinking that the backend boundary needs to be bytes (so, `Vec<u8>`/`&[u8]`/`Bytes`) and nargo serializes it and a backend deserializes it using the ACVM crate, so I was pretty sure we'd need `read` and `write` functions

TomAFrench: I don't follow. This serialisation method is only used when saving the witness to disk rather than when communicating with backends.

acir/src/native_types/witness.rs
TomAFrench: Does `into_values` give us any more specific insight into the underlying implementation than this function though? They're the same thing but with a `.collect()`.

phated: It kind of implies there are 2 types of iterators; otherwise you'd just implement `Iter`. I thought about implementing `Iter` and giving it a tuple of `(Witness, FieldElement)`—would you want that instead?

TomAFrench: Yeah, I think returning an iterator over key-value pairs is the least opinionated iterator to give.

phated: Done, since it also needed to export the `IntoIter` struct, I moved it to a separate module. I'm in the process of updating the aztec_backend and Noir.

acir/src/native_types/witness.rs
phated: Done and rebased on master

phated: Let's keep these for now since removing them (and the follow up discussion) is unrelated to this PR

TomAFrench (approved):
LGTM! Seeing as I touched this last I'll let you merge into master.

TomAFrench:
Updating the downstream PRs are blocked on https://github.com/noir-lang/aztec_backend/pull/151

PR #257, opened 2023-05-03
feat!: Remove `solve` from PWG trait & introduce separate solvers for each blackbox

phated:
# Related issue(s)

Works towards a resolution of the conversation at https://github.com/noir-lang/acvm/pull/247/files#r1180617953 and puts ACVM in a better position for self-implementing all of the opcodes (one by one).

# Description

## Summary of changes

This removes the `solve` function on the PartialWitnessGenerator trait and moves it into the `pwg` module as a standalone function that takes a backend. It also replaces `solve_black_box_function_call` with individual functions for each black box function being solved. This will allow us to systematically implement the blackbox function calls in ACVM and then remove them from the PWG interface for backends. Until (eventually) they are all gone and the backend doesn't need to implement anything for PWG; instead, just conforming to the ACVM implementation.

I've split these out into separate functions because leaking the BlackBoxFunc enum to a backend is misleading after #247 lands and keccak is handled by ACVM. However, it seems like we'll need pedersen for quite some time (probably as the only one?).

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -7,16 +7,14 @@
 pub mod compiler;
 pub mod pwg;
 
-use crate::pwg::{arithmetic::ArithmeticSolver, oracle::OracleSolver};
 use acir::{
     circuit::{
-        opcodes::{BlackBoxFuncCall, OracleData},
+        opcodes::{BlackBoxFuncCall, FunctionInput},
         Circuit, Opcode,
     },
     native_types::{Expression, Witness},
     BlackBoxFunc,
 };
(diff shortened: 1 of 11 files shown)


phated:
@guipublic reviewed via DM while github was broken, I've pushed his suggested changes, so should be good for another review.

guipublic (approved):

TomAFrench:
Much prefer this compared to original implementation as we can enforce removal of dead code in backends by removing bb functions from the trait.

PR #264, opened 2023-05-05
feat!: Remove backend solvable methods from the interface and solve them in ACVM

guipublic:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

Use ACVM instead of calling the backend for solving blackbox functions, for the bb functions which do not use barretenberg for solving (the low hanging fruits).

# Checklist

# Additional context

Once this PR is merged, I will remove the related solve traits from pwg so that backends do not need to implement these solvers.


acvm/src/pwg/blackbox.rs
```diff
@@ -39,56 +45,36 @@ pub(crate) fn solve(
...
+        BlackBoxFunc::HashToField128Security => {
+            hash_to_field_128_security(initial_witness, bb_func)
         }
-        BlackBoxFuncCall { name: BlackBoxFunc::Keccak256, inputs, outputs } => {
-            backend.keccak256(initial_witness, inputs, outputs)
+        BlackBoxFunc::EcdsaSecp256k1 => secp256k1_prehashed(initial_witness, bb_func),
+        BlackBoxFunc::FixedBaseScalarMul => {
+            backend.fixed_base_scalar_mul(initial_witness, &bb_func.inputs, &bb_func.outputs)
         }
+        BlackBoxFunc::AES => Err(OpcodeResolutionError::UnsupportedBlackBoxFunc(bb_func.name)),
```
TomAFrench, line 78: Why do we block this at the ACVM level?

acvm/src/pwg/blackbox.rs
```diff
@@ -39,56 +45,36 @@ pub(crate) fn solve(
...
+            backend.pedersen(initial_witness, &bb_func.inputs, &bb_func.outputs)
         }
-        BlackBoxFuncCall { name: BlackBoxFunc::FixedBaseScalarMul, inputs, outputs } => {
-            backend.fixed_base_scalar_mul(initial_witness, inputs, outputs)
+        BlackBoxFunc::HashToField128Security => {
+            hash_to_field_128_security(initial_witness, bb_func)
         }
-        BlackBoxFuncCall { name: BlackBoxFunc::Keccak256, inputs, outputs } => {
-            backend.keccak256(initial_witness, inputs, outputs)
+        BlackBoxFunc::EcdsaSecp256k1 => secp256k1_prehashed(initial_witness, bb_func),
```
TomAFrench, line 74: This should live next to the other ACVM implemented functions. I'd like to keep these separated from each other.

TomAFrench (commented):
Much happier with this compared to original proposal as now we've defined a proper interface. This PR is leaving a lot of dead code on the PWG trait however which needs to be cleaned up.

acvm/src/pwg/blackbox.rs
guipublic: Indeed it does not make sense here, if it is not supported in ACVM, we should not have an opcode for it!
I will call the backend instead. However I suppose it should be easy to implement in ACVM, but it can be done in another PR. Do we have a specification for this opcode?

acvm/src/pwg/blackbox.rs
guipublic: I am not sure what you mean, could you clarify your comment?

guipublic: Done

TomAFrench:
This PR is still leaving dead code on the `PartialWitnessGenerator` trait.

guipublic:
Yes, the idea is to remove this in another PR when this one is merged, we will be able to remove corresponding solves in the backend safely since they will not be called after this PR is merged.

phated:
I think we could rename the PR to `feat!: Solve blackbox functions in ACVM instead of backends` and then remove the functions from the PWG trait in this same PR. It'd communicate to implementors accurately, as this `chore` would be excluded from the changelog currently.

guipublic:
Since we want to have small PRs, I think it makes sense to have this one which does not call the backend and implement the solve methods, but does not break anything (and so it is a chore), and then have another one which will be a feat! and will break the interface, but it will only do that. This way there is no confusion between implementing the solvers and modifying the PWG trait.

TomAFrench:
Agreed with renaming PR, I don't really see the logic in doing this in two steps. We should only have one implementation of the black box functions in use at any one time.

phated:
I don't think this PR is non-breaking. If a backend was relying on their implementation of `backend.schnorr_verify` but you switch it in this PR, then they are broken. If we remove the PWG functions in the same PR, it is very clear that there isn't 2 separate implementations.

guipublic:
> If a backend was relying on their implementation of backend.schnorr_verify but you switch it in this PR, then they are broken

No they are not be broken if I switch-in this PR, the schnorr verification should work for both implementations, if not that means their implementation was already broken.

TomAFrench:
Another example would be `HashToField128Security`, if a backend were using a hash function other than blake2s then this would be breaking.

kevaundray:
If a backend was not doing the _correct_ behavior for functions like hash_to_field etc, then I think thats on them -- and us for not having a spec.

I'm impartial as to this PR being breaking, but we should not release a version without the accompanying breaking change which removes the duplicates.

acvm/src/pwg/blackbox.rs
kevaundray: Might make sense to remove it in another PR until we have proper support for it, it could be aes128 or aes256

kevaundray (approved):

PR #258, opened 2023-05-04
chore!: backend can specify which opcode they support

guipublic:
# Description

## Summary of changes

replace blackbox_supported with opcode_supported

## Dependency additions / changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

- [ ] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` with default settings.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this PR to the issue(s) that it resolves.
- [X] I have reviewed the changes on GitHub, line by line.
- [X] I have ensured all changes are covered in the description.

# Additional context

This is useful for memory opcodes, we could have a TP backend which supports only ROM opcode for instance.

PR #269, opened 2023-05-08
feat!: use struct variants for blackbox function calls

sirasistant:
# Related issue(s)

# Description

## Summary of changes
Define different types for the variants of BB function call
## Test additions / changes

# Checklist

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` with default settings.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this PR to the issue(s) that it resolves.
- [ ] I have reviewed the changes on GitHub, line by line.
- [ ] I have ensured all changes are covered in the description.

# Additional context

(If applicable.)

PR #273, opened 2023-05-09
chore!: allow backends to specify support for all opcode variants

kevaundray:
Related to #258 

# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)


kevaundray:
@TomAFrench See the diff for adding Fn trait

acvm/src/lib.rs
```diff
@@ -159,8 +159,8 @@ pub trait ProofSystemCompiler {
     /// if the language and proof system does not line up.
     fn np_language(&self) -> Language;
 
-    // Returns true if the backend supports the selected black box function
-    fn black_box_function_supported(&self, opcode: &BlackBoxFunc) -> bool;
+    // Returns true if the backend supports the selected opcode
+    fn opcode_supported(&self, opcode: &Opcode) -> bool;
```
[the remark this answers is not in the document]
TomAFrench: We can create a stubbed `ProofSystemCompiler` here in a test similar how we do for `PartialWitnessGenerator`.

It doesn't match by itself but we can wrap it in a closure.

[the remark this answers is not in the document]
kevaundray: The function trait is `Fn` trait so the &self should not matter here -- I can quickly add an example backend in a test to show for the PR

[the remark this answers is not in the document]
kevaundray: Added a test with a dummy compiler that calls the compile method -- its very verbose though, so I'd not want to keep it since its testing compile time behaviour.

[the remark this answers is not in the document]
kevaundray: Didn't see @TomAFrench reply until I refreshed the page :)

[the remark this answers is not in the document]
phated: @kevaundray No worries. A lot of my experiments with pluggable backends also remove `self` so now I have an extra reason to remove them in the future 😛

acvm/src/compiler.rs
```diff
@@ -106,3 +106,61 @@ pub fn compile(
...
 }
+
+#[test]
+fn compile_api_compiles_with_backend() {
+    use crate::ProofSystemCompiler;
+
+    compile(
+        Circuit::default(),
+        Language::R1CS,
+        |opcode| DummyProofSystem.opcode_supported(opcode),
```
phated, line 117: So is this never called because it is R1CS? I was confused when `opcode_supported` has a panic but it is used here.

TomAFrench: This is never called as the `Circuit` doesn't contain any opcodes to check.

acvm/src/lib.rs
```diff
@@ -159,8 +159,8 @@ pub trait ProofSystemCompiler {
     /// if the language and proof system does not line up.
     fn np_language(&self) -> Language;
 
-    // Returns true if the backend supports the selected black box function
-    fn black_box_function_supported(&self, opcode: &BlackBoxFunc) -> bool;
```
TomAFrench, line 163: I think we may want to retain something which allows a frontend to see which black box functions a backend supports without having to construct a full opcode. This is relevant for #269.

Alternatively we need to formalise some way to create dummy opcodes to test compatibility: `Opcode::BlackBoxFuncCall(BlackBoxFuncCall::dummy(BlackBoxFunc))` 

This is also needed for other opcodes where we want to check support where we don't have a concrete opcode yet. https://github.com/noir-lang/noir/pull/1282/files#diff-4b6abdc6086c59b7255bb6d107f5bc0cec0c68b18a26b0eaebad9671ee5b570fR139

acvm/src/lib.rs
```diff
@@ -227,9 +227,7 @@ pub fn checksum_constraint_system(cs: &Circuit) -> u32 {
 // This is set to match the previous functionality that we had
 // Where we could deduce what opcodes were supported
 // by knowing the np complete language
-pub fn default_is_opcode_supported(
-    language: Language,
-) -> compiler::transformers::IsOpcodeSupported {
+pub fn default_is_opcode_supported(language: Language) -> impl Fn(&Opcode) -> bool {
```
TomAFrench, line 230:
```suggestion
pub fn default_is_opcode_supported(language: Language) -> fn(&Opcode) -> bool {
```

We can make this type stricter as we're not going to be capturing the environment here.

acvm/src/compiler.rs
```diff
@@ -106,3 +106,61 @@ pub fn compile(
         return_values: acir.return_values,
     })
 }
+
+#[test]
+fn compile_api_compiles_with_backend() {
```
TomAFrench, line 111: I think we can remove this test case as it's just to show that we can pass in a closure here.

TomAFrench (approved):
I think this is good to merge at this point. Some nits I don't feel too strongly about.

acvm/src/lib.rs
TomAFrench: Made this change and tested that this is still accepted as a `impl Fn(&Opcode) -> bool` as a sanity check.

TomAFrench:
I touched this last so will let one of you two have the last word before merging.

sirasistant (approved):
LGTM

PR #281, opened 2023-05-11
feat!: added hash index to pedersen

sirasistant:
# Related issue(s)
Related issue: https://github.com/noir-lang/noir/issues/1341
# Description

Adds the hash index constant parameter to pedersen bb func call.

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)


guipublic (commented):
What this is doing?
Do we have a corresponding GitHub issue to motivate the change?

TomAFrench:
Same, I'm missing a lot of context around this change so it would be helpful to have a full issue on what this is trying to solve.

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -62,6 +62,7 @@ pub enum BlackBoxFuncCall {
     },
     Pedersen {
         inputs: Vec<FunctionInput>,
+        hash_index: u32,
```
kevaundray, line 65: Perhaps this is better called as `domain_separator`? Since it is an index that will be used to separate different hash calls

sirasistant:
Related issue: https://github.com/noir-lang/noir/issues/1341

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -366,7 +373,11 @@ impl BlackBoxFuncCall {
                     })
                 }
             }
-            BlackBoxFunc::Pedersen => Ok(BlackBoxFuncCall::Pedersen { inputs, outputs }),
+            BlackBoxFunc::Pedersen => Ok(BlackBoxFuncCall::Pedersen {
+                inputs,
+                domain_separator: read_u32(&mut reader)?,
```
guipublic, line 378: I find the mixing of read with contructing the object messy.
This would be better with a common read input-outputs function and dedicated read/write functions per blackbox which can call the read input/outputs if needed.

guipublic (dismissed):
This new field indicates which set of generators to use. The fact that we use it in Aztec for domain separation is not relevant in this context. A better name would be something like generators_set or generators_index.

kevaundray:
> This new field indicates which set of generators to use. The fact that we use it in Aztec for domain separation is not relevant in this context. A better name would be something like generators_set or generators_index.

Its main purpose is domain separation, so this name is fine

kevaundray:
@guipublic 

The naming of generator_set vs domain_separation is also related to #282 ;

If we had a homomorphic pedersen implementation, then generator_set would be the correct choice, but since we are using it as a HashToGroup, domain_separation works better here.

guipublic:
In any cases (homomorphic or not) it indicates which group elements to use, so generator_set is ok.
When the opcode will be renamed, then maybe another name will be better but we can (and should) change it at that time, not now.

kevaundray:
> In any cases (homomorphic or not) it indicates which group elements to use, so generator_set is ok.
> When the opcode will be renamed, then maybe another name will be better but we can (and should) change it at that time, not now.

generator_set is not a detail that matters here, what we need is domain separation and knowing what the generator_set is, does not help in anyway due to this function not being homomorphic right now. 

Lets go with `domain_separator`and we can come back to it, if it proves to not be sufficient.

guipublic:
> generator_set is not a detail that matters here, what we need is domain separation 

generator_set is a detail that does not matter for Aztec, what we need for Aztec is domain separation. When we will set this value coming from a noir contract, it will be called domain_index or domain_separator.

But at the acir level, domain_separator for a pedersen opcode is meaningless.

>and knowing what the generator_set is, does not help in anyway due to this function not being homomorphic right now.

Knowing what the generators are allows someone to know what this opcode is doing, not what is his purpose, but what exact computation it is proving.

acir/src/circuit/opcodes/black_box_function_call.rs
sirasistant: But since they are variants of the same enum the typing on the variant specifc functions would be pretty bad:
 ```rust
 fn read_aes<R: Read>(mut reader: R) -> std::io::Result<BlackBoxFuncCall> {
    let inputs = read_inputs(&mut reader)?;
    let outputs = read_outputs(&mut reader)?;
    Ok(BlackBoxFuncCall::AES { inputs, outputs })
}
 
 fn write_aes<W: Write>(aes: &BlackBoxFuncCall, mut writer: W) -> std::io::Result<()> {
    if let BlackBoxFuncCall::AES { inputs, outputs } = aes {
        write_inputs(inputs, &mut writer)?;
        write_outputs(outputs, &mut writer)?;
        Ok(())
    } else {
        Err(std::io::ErrorKind::InvalidData.into())
    }
}

 ```

guipublic: I think it works for the read(), but the write is not nice indeed.
What do you think of having the function for the read and keep the write as you did?

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -300,101 +438,33 @@ impl BlackBoxFuncCall {
         write_inputs(&self.get_inputs_vec(), &mut writer)?;
         write_outputs(&self.get_outputs_vec(), &mut writer)?;
 
-        Ok(())
+        match self {
+            BlackBoxFuncCall::Pedersen { domain_separator, .. } => {
+                write_u32(&mut writer, *domain_separator)?;
```
vezenovm, line 443: The `domain_separator` coming after the outputs feels strange. Will it not be an input set by the user?

vezenovm: If we move forward with custom methods for reading in the BB calls we should do the same for writing them. However, this will require us to match inside of the `write` function like we do for a `read`. This can be avoided though by using the strategy taken in this PR: https://github.com/noir-lang/acvm/pull/285/files

sirasistant: It's a little different in the sense that it's a constant value, not a witness index, but yeah conceptually is also an input! Seeing that now some functions might have multiple variable length arrays of witness indices as inputs, I think I would go with fully custom (de)serialization methods per bb call yep

vezenovm: We can avoid having custom write method using the strategy in PR #285 it seems though. Right now we are going to need a match statement in `write` like we have in `read`

vezenovm: > constant value, not a witness index, but yeah conceptually is also an input

Like you said constants are still inputs and should encapsulated as part of the writing of the inputs imo

sirasistant: But in that case we'd be serializing unnecesary length statements when the arguments are known-length, just to avoid, i feel that the option of specific write functions is more expressive. And if the match inside each write function is a problem (because it's a runtime check) we can type the content of each enum variant as a different struct, and provide a read/write trait impl for the structs that represent the data of each struct variant!

vezenovm: I would want to get rid of the internal write method match statements. However, I don't think it is possible in Rust atm to impl on enum variants. We would have to impl on the enum itself. Which at that point requires us to match on the whole enum.

I'm going to try and see how I can avoid unnecessary length statement's by differentiating the function inputs with an enum

kevaundray (approved):

PR #284, opened 2023-05-16
feat(acvm)!: expose backend identifier on `ProofSystemCompiler` trait

### LINKED ISSUE #196: Expose backend identifier information
### Problem

We currently have no way for nargo to actually know what backend it's interacting with. An example of where this is useful is [embedding this into preprocessed artifacts](https://github.com/noir-lang/noir/pull/1126#discussion_r1164092913) as these are backend specific so it's important to know which backend to use.

### Proposed solution

We should add a function which exposes useful information for callers. This could be as simple as exposing the backend identifier but we may want to expose more information such as turbo vs ultraplonk, etc.

### Alternatives considered

_No response_

### Additional context

_No response_

### Submission Checklist

TomAFrench:
# Related issue(s)

Resolves #196 

# Description

## Summary of changes

This PR exposes the `backend_identifer` method on `ProofSystemCompiler` which returns a string identifier for the backend.

I'm doing an initial pass through the repos to just include the name of the backend. We can add extra info later on before merging.

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -65,13 +65,19 @@ pub trait Backend:
 {
 }
 
+// TODO: improve this name
+pub trait BackendInfo {
+    /// Returns the identifier for the backend.
+    fn identifier(&self) -> String;
+}
+
 // Unfortunately, Rust doesn't natively allow async functions in traits yet.
 // So we need to annotate our trait with this macro and backends need to attach the macro to their `impl`.
 //
 // For more details, see https://docs.rs/async-trait/latest/async_trait/
 // and https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/
 #[async_trait]
-pub trait CommonReferenceString {
+pub trait CommonReferenceString: BackendInfo {
     /// The Error type returned by failed function calls in the CommonReferenceString trait.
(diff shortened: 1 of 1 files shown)


acvm/src/lib.rs
```diff
@@ -207,6 +207,9 @@ pub trait ProofSystemCompiler {
     /// The Error type returned by failed function calls in the ProofSystemCompiler trait.
     type Error: std::error::Error; // fully-qualified named because thiserror is `use`d at the top of the crate
 
+    /// Returns the identifier for the backend.
+    fn backend_identifier(&self) -> String;
```
phated, line 211: This is also needed by code that calls the CommonReferenceString trait, so do we want it to be on the Backend trait itself or a separate trait that others can depend upon?

TomAFrench: Hmm, I don't really want to put this on `Backend` as that would hamper our ability to pass `&impl SubTraits` around effectively.

I placed it on `ProofSystemCompiler` as this is the trait which is most tightly bound to the underlying proving library. `PWG`/`SmartContract` shouldn't ever need to expose this information so we could have a separate trait which is just `ProofSystemCompiler` and `CommonReferenceString` depend on.

TomAFrench:
This trait is being removed

Issue #228, opened 2023-04-25
[Bug/Security] Out-of-memory (OOM) panic while reading a circuit (MemoryBlock)


pventuzelo:
### Aim

We (@fuzzinglabs & @lambdaclass) found an out-of-memory issue in acvm 0.9.0 while reading a circuit.

### Expected behavior

It should not crash.

## Root cause

```
memory allocation of 167607452400 bytes failed
[1]    3680483 abort (core dumped
```

Related code: https://github.com/noir-lang/acvm/blob/7352802ca8c4abc9be21ac4f88c8c6b7064d9650/acir/src/circuit/opcodes.rs#L41

This issue occurs when reading a MemoryBlock. Before calling `Vec::with_capacity`, the `len` value should be checked and sanitized.

This issue is similar to https://github.com/noir-lang/acvm/issues/187 and a similar fix using `try_reserve_exact` can be used here.

## Detailed behavior
``` sh
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c4c859 in __GI_abort () at abort.c:79
#2  0x00005555555ac4d7 in std::sys::unix::abort_internal () at library/std/src/sys/unix/mod.rs:350
#3  0x000055555555ddd7 in std::process::abort () at library/std/src/process.rs:2138
#4  0x00005555555ac94e in std::alloc::rust_oom () at library/std/src/alloc.rs:356
#5  0x00005555555ac957 in std::alloc::_::__rg_oom () at library/std/src/alloc.rs:351
#6  0x00005555555c0f47 in alloc::alloc::handle_alloc_error::rt_error () at library/alloc/src/alloc.rs:385
#7  0x000055555555eec7 in alloc::alloc::handle_alloc_error () at library/alloc/src/alloc.rs:389
#8  0x0000555555560589 in alloc::raw_vec::RawVec<T,A>::allocate_in (capacity=698364385, init=alloc::raw_vec::AllocInit::Uninitialized, alloc=...) at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/alloc/src/raw_vec.rs:190
#9  0x000055555556a7e7 in alloc::raw_vec::RawVec<T,A>::with_capacity_in () at /rustc/a266f11990d9544ee408e213e1eec8cc9eb032b7/library/alloc/src/raw_vec.rs:131
[...]

PR #286, opened 2023-05-16
chore!: Remove manual serialization of `Opcode`s in favour of `serde`

kevaundray:
# Related issue(s)

Resolves (link to issue)

# Description

Discussed internally: We originally wanted an encoding which could be implemented in other languages easier. I thought that having a manual encoding and then specifying this in some documentation would have been the best way for this. Even if this was the case, we have had two serialization bugs and the effort for folks to modify this code has become unproductive.

There are still solutions to the problem of having other languages implement a deserialization strategy, they can choose to use rmp/json for example, which is available in most languages. For now, we use rmp by default with the read and write methods. 

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)


kevaundray:
Closes #228

acir/src/circuit/opcodes.rs
```diff
@@ -224,44 +154,3 @@ impl std::fmt::Debug for Opcode {
         std::fmt::Display::fmt(self, f)
     }
 }
-
-#[test]
-fn serialization_roundtrip() {
```
vezenovm, line 229: Couldn't hurt to still have the serialization_roundtrip test? At least here for the main circuit where we still have `read` and `write`

kevaundray: I left it in for the main Circuit in `acir/src/circuit/mod.rs` -- The derive(Serialize, Deserialize) should take care of it for opcodes though

vezenovm: Oops I missed that I thought you deleted it. Perhaps we should copy-paste some of the directives, opcodes, etc. from the other serialization tests into the test in `acir/src/circuit/mod.rs` for completeness sake

kevaundray: Yep makes sense

vezenovm:
I think you can remove `to_u16()` and `from_u16()` in black box func file

kevaundray:
> I think you can remove `to_u16()` and `from_u16()` in black box func file

This was there prior to the manual serialization work so we probably want to check if this is being used in other places.

vezenovm:
> This was there prior to the manual serialization work so we probably want to check if this is being used in other places.

I see it only being used in the `consistent_index` test in `black_box_functions.rs`

TomAFrench (approved):

PR #291, opened 2023-05-17
feat(acir)!: Verify Proof

### LINKED ISSUE #198: Add recursion opcode

vezenovm:
# Related issue(s)

Resolves #198 

The `acvm-backend-barretenberg` PR which implements this can be found here:
https://github.com/noir-lang/acvm-backend-barretenberg/pull/207
The relevant Noir PR which implements this and the backend PR can be found here:
https://github.com/noir-lang/noir/pull/1491

# Description

Add VerifyProof variant to BlackBoxFunc enum

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)


acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -192,6 +211,32 @@ impl BlackBoxFuncCall {
...
+                        + 1
+                        + input_aggregation_object.len(),
+                );
+                inputs.extend(key.iter().copied());
+                inputs.extend(proof.iter().copied());
+                inputs.extend(public_inputs.iter().copied());
+                inputs.push(*key_hash);
+                // If we do not have an input aggregation object assigned
+                // do not return it as part of the input vector as to not trigger
+                if !input_aggregation_object.iter().any(|v| v.witness == Witness(0)) {
```
kevaundray, line 235: Witness(0) seems to be a barretenberg specific idiom?

acvm/src/lib.rs
```diff
@@ -150,6 +159,7 @@ pub trait ProofSystemCompiler {
         circuit: &Circuit,
         witness_values: WitnessMap,
         proving_key: &[u8],
+        is_recursive: bool,
```
kevaundray, line 162: Why is this needed and can it not be deduced from the circuit?

acvm/src/lib.rs
```diff
@@ -160,5 +170,18 @@ pub trait ProofSystemCompiler {
         public_inputs: WitnessMap,
         circuit: &Circuit,
         verification_key: &[u8],
+        is_recursive: bool,
     ) -> Result<bool, Self::Error>;
+
+    fn proof_as_fields(
```
kevaundray, line 176: Why is this and the below function needed?

vezenovm: We need a proof and verification key to be witnesses of the circuit in order to recursively verify them. Most proofs and verification key's are not formatted in a proof system's native field elements. I was debating to put this in `acvm-backend-barretenberg` but it felt universal enough for being able to perform recursion with Noir in general. 

@zac-williamson do you think these methods are too specific for the acvm and should possibly just be moved into our backend? For example, if a halo2 backend were to implement recursion do you think these methods would still be necessary?

acvm/src/lib.rs
vezenovm: Unfortunately it cannot really be deduced from the circuit. An inner circuit may require a different prover in order to be verified recursively (in the case of bberg the inner circuit prover must be based upon Pedersen-Blake3 rather than Keccak). The internal circuit is able to just be a normal circuit that looks no different from any other Noir circuit

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: So I was debating how to go about this here. I could add a flag that determines whether or not we should include the `input_aggregation_object` in our inputs vector instead of this check. If the `input_aggregation_object` is all zeros we will get a missing assignment error when we call `solve` in `blackbox.rs`. Setting the `input_aggregation_object` as all zeros is specific to our bberg recursion circuit, but either way we are going to have to some sort of indicator whether to have an input aggregation object assigned. I was thinking a flag is similar to requiring anyone who is integrating the VerifyProof blackbox func to have the `input_aggregation_object` of their first recursive verification be all zeros. However, a flag is more explicit and I can switch to this

acvm/src/lib.rs
kevaundray: what does is_recursive denote/mean?

acvm/src/lib.rs
kevaundray: How/Where would these two methods be called?

acir/src/circuit/opcodes/black_box_function_call.rs
kevaundray: We don't have anything in acvm right now that states that Witness(0) is special

Can we have `input_aggregation_object` be optional then explain what it means for it to be None?

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -76,6 +76,16 @@ pub enum BlackBoxFuncCall {
         inputs: Vec<FunctionInput>,
         outputs: Vec<Witness>,
     },
+    VerifyProof {
+        key: Vec<FunctionInput>,
```
kevaundray, line 80:
```suggestion
        verification_key: Vec<FunctionInput>,
```

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -76,6 +76,16 @@ pub enum BlackBoxFuncCall {
...
+        proof: Vec<FunctionInput>,
+        public_inputs: Vec<FunctionInput>,
```
kevaundray, line 82: Add a comment here specifying what these `public_inputs` refer to

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -76,6 +76,16 @@ pub enum BlackBoxFuncCall {
...
+        key_hash: FunctionInput,
```
kevaundray, line 83: Add comment specifying whether this is the proving/verification key hash and also why it is needed

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -76,6 +76,16 @@ pub enum BlackBoxFuncCall {
         inputs: Vec<FunctionInput>,
         outputs: Vec<Witness>,
     },
+    VerifyProof {
```
kevaundray, line 79: I don't think VerifyProof is the right name for this opcode, since that would indicate that this should return a boolean as output and or at the least that once this opcode was called, that the proof was verified

acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -76,6 +76,16 @@ pub enum BlackBoxFuncCall {
...
+        key: Vec<FunctionInput>,
+        proof: Vec<FunctionInput>,
+        public_inputs: Vec<FunctionInput>,
+        key_hash: FunctionInput,
+        input_aggregation_object: Vec<FunctionInput>,
+        // This is the recursive verification output aggregation object.
+        // The name `outputs` was kept to simplify code reuse with the other BlackBoxFuncCall's
```
kevaundray, line 86: I think changing this to output_aggregation_object would be better especially since we no longer have a homogenous API for the black box functions

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: Hmm yes good point. We could call it `VerifyRecursiveProof` or `RecursiveVerification` (although I do `VerifyProof` implies this as well)? I will leave comments indicating the purpose of the input and output aggregation object

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: I will change it to the specific name

acir/src/circuit/opcodes/black_box_function_call.rs
kevaundray: Hmm it seems to be aggregating and not entirely verifying

kevaundray (commented):
Generally looks okay -- Some overall questions:

- Can you give a brief outline as to how this opcode will be used from the user perspective? (I'm trying to understand the rationale behind this interface)

- The size of the aggregation objects can change depending on the proving system, where do we account for this?

acvm/src/lib.rs
vezenovm: These would be called by someone looking to pass a proof and/or verification key into their Noir circuit. The way our recursive verification opcode is setup we require a verification key and a proof and these must be represented by witness values. 

Any backend is going to have its own recursion format for a proof and/or vkeys. Being able to export the proof and verification key in the appropriate format will be necessary for any backend looking to integrate recursion. I could move this functionality to acvm-backend-bberg but then I will still want to expose functionality for a backend to be able to generate the appropriate recursive format for a proof and vkey. 

Having these methods be a part of the ACVM allows for us to more easily integrate it into `nargo`. Otherwise this will be a purely backend specific functionality, when it really is not. 

Let's say I have circuit A. I compile and generate a proof for this circuit.

I want to verify circuit A in circuit B. I need to pass in the proof and vkey in the appropriate recursion format. These methods will be used to generate the proof and vkey as fields (perhaps I should rename the functions `***_in_recursion_format`?). The outputted proof and vkey as fields will be input to circuit B and used as parameters for the verify proof builtin.

acvm/src/lib.rs
vezenovm: It means we want to generate a circuit which will be recursively verified in another circuit. But how it is being used right now is essentially to determine the hash type of the prover for an inner circuit. I do agree though this is not general enough and I will consider if there is a better way to abstract this functionality. What are your thoughts on a `hash_type` flag that determines the fiat shamir hash used by the prover?

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: We do verify the previous aggregation state technically, and then aggregate the current proof onto it. So aggregation may be a better name. `RecursiveAggregation`?

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: Yes I can switch to that instead

acvm/src/lib.rs
vezenovm: So the inner proof that is generated may use a different algorithm than a normal proof that can be verified by the Solidity verifier. The inner proof in this case may use a different hash function for Fiat-Shamir as that hash is more snark friendly and more optimized for recursive verification. 

This is information that would be passed to the proof system, however, there is no way to indicate from the circuit alone that we want to use a different hash type for the prover. I think if the `is_recursive` flag is too specific (and I agree is a bit of a leakage from bberg) we will need some other flag. This flag can either be a `hash_type` flag as proposed above or a custom data index of some sort. 

Perhaps a custom_data index of type [u8] would be simplest and then we just leave it for whichever backend is implementing these methods to indicate what custom data means. I imagine we would then want the backend to expose some sort of custom_data API as nargo is going to have to integrate these methods

kevaundray: Its mainly because when you create proofs that you want to be natively verfied, you use perhaps sha256 for your hash function in the proof system. But when you want to verify the proof in a snark, you want the hash function being used to be snark friendly for the fiat-shamir/transcript. is_recursive will use a snark-friendly hash function when true and otherwise it will use the regular one, which for bb is keccak, but it can be anything for a proving system

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: I switched to an Option. It is now entirely up to a backend to determine how they will handle the first recursive aggregation

acvm/src/lib.rs
vezenovm: I added a comment on this function. It might not be descriptive enough so let me know what you think

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: changed

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: Comments added

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: Commented

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: I changed the name to `RecursiveAggregation`

acir/src/circuit/opcodes/black_box_function_call.rs
vezenovm: Changed

acir/src/circuit/black_box_functions.rs
```diff
@@ -44,6 +45,7 @@ impl BlackBoxFunc {
             BlackBoxFunc::XOR => "xor",
             BlackBoxFunc::RANGE => "range",
             BlackBoxFunc::Keccak256 => "keccak256",
+            BlackBoxFunc::VerifyProof => "verify_proof",
```
vezenovm, line 48: I changed the name of the `BlackBoxFuncCall` to `RecursiveAggregation` but I was unsure on changing the `BlackBoxFunc` call to that as well. Leaving the builtin with the name `verify_proof` I feel like would still be good but then the mismatch with `BlackBoxFuncCall::RecursiveAggregation` is a bit deceiving.

acvm/src/lib.rs
vezenovm: I added some comments to explain `is_recursive`

acvm/src/lib.rs
```diff
@@ -101,6 +101,15 @@ pub trait PartialWitnessGenerator {
         input: &FunctionInput,
         outputs: &[Witness],
     ) -> Result<OpcodeResolution, OpcodeResolutionError>;
+    fn verify_proof(
```
kevaundray, line 104: I forget the conclusion of our discussion re having this done inside of the backend. 

I'll information dump for context.

Removing it gives an imbalance because the RecursiveAggregation opcode will be the only one without a solver method.

Keeping it means that the solver will either directly or indirectly need a backend (unless the solver implements it to return true) + it will be a component that will now need to be setted when we attempt to solve.

The root root problem being that there is no standard way to do recursive aggregation, so its not possible to implement it here like the other opcodes. However, since these witnesses are never used by the business logic in the program (fed into another opcode) we can optimistically assume that it verifies true and ignore it.

The conversation over slack was that we needed to set the output values `output_aggregation_object`using this method, but it was noted that we do not need to set these until we get to the backend since they are never used in other places in the IR, so we could remove this from the ACVM layer altogether.

vezenovm: I think we are safe to remove this from the ACVM layer. However, it does feel weird for `RecursiveAggregation` to not have a solver method. I can remove this from the ACVM for now and leave a comment describing why it is unnecessary and our choice to always have the recursion opcode pass execution during simulation.

acir/src/circuit/black_box_functions.rs
kevaundray: Yeah its somewhat confusing I think, especially since its not verifying a proof

kevaundray (approved):

PR #302, opened 2023-05-22
chore!: Remove AES opcode

kevaundray:
# Related issue(s)

Resolves (link to issue)

# Description

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/black_box_functions.rs
+++ b/acir/src/circuit/black_box_functions.rs
@@ -7,7 +7,7 @@ use strum_macros::EnumIter;
 #[cfg_attr(test, derive(EnumIter))]
 pub enum BlackBoxFunc {
     #[allow(clippy::upper_case_acronyms)]
-    AES,
+    AES128,
     AND,
     XOR,
     RANGE,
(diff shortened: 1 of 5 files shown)


acir/src/circuit/black_box_functions.rs
```diff
@@ -7,7 +7,7 @@ use strum_macros::EnumIter;
 #[cfg_attr(test, derive(EnumIter))]
 pub enum BlackBoxFunc {
     #[allow(clippy::upper_case_acronyms)]
-    AES,
+    AES128,
```
TomAFrench, line 10: We don't have any specification for this opcode though do we? i.e. we're going to need to provide a message + private key but how are they laid out in the list of function inputs?

Also is this AES encryption or decryption? Should we have two opcodes or have a extra input to choose between? (likely the former)

kevaundray: Good points -- I'm gonna block this PR until we can answer those, alternatively we can remove the AES opcode and when we have a proper description for it, we can add it back. What do you think?

TomAFrench: I think if we don't have a good idea of what it should do then we should remove it. Do we have something in Barretenberg which we can hook this up to?

kevaundray: I'm okay with removing it for now, we do have a function in barretenberg but it has not seen much usage so it could have a bunch of bugs

PR #314, opened 2023-05-24
feat!: Add variable length keccak opcode

### LINKED ISSUE #313: Add Keccak Opcode which allows the message size to vary at prover time
### Problem

For most of our hash functions, we have the property that the length of the message needs to be known at compile time.

We have recently discovered that this usecase is not sufficient for real-world applications like RLP. 

The difference here being that the size of the message may be known by the prover.

### Happy Case

The solution to this is to introduce an extra parameter to Keccak, lets name it **actual_length** which is a witness and determines the actual length of the message.

Lets discuss the idiosyncranicities of this. 

## Before this proposal

We can do the following:

```
let hash_a = keccak([1,2,3,4])
```

```
let hash_b = keccak([1,2,3,4,0,0,0,0])
```

The second function is padded with four zeroes, while the first is not. It is evident that `hash_a` and `hash_b` are not the same due to the collision resistance property of keccak.

Given a circuit, it is not possible to produce `hash_a` and `hash_b` because the message size needs to be fixed.

## After this proposal

We can do the following:

```
let message_size = 4;
let hash_a = keccak([1,2,3,4], message_size)
```

```
let message_size = 4;
let hash_b = keccak([1,2,3,4,0,0,0,0], message_size)
```

`hash_a` and `hash_b` will produce the same value. This is because the `message_size` indicates how many bytes in the message we should actually be hashing. 

Note that even though `hash_a` and `hash_b` produce the same hash.  `hash_b`  is more expensive because the fixed message_size is 8 bytes. This is due to us passing `[1,2,3,4,0,0,0,0]`.

The hash function can then be seen to have two configuration parameters:

- MAX_MESSAGE_SIZE : This is a circuit constant.
- ACTUAL_MESSAGE_SIZE: This is a prover provided value, that must be less than MAX_MESSAGE_SIZE. One can separately add a constraint to make this equal to a constant, so that it has the same behaviour as the previous Keccak method.

## Things to Note

- Keccak with this extra argument, will always be more expensive than the Keccak that only allows constant messages. For this reason, it makes sense to use a separate opcode, If the backend wants to under the hood wants to use one implementation, then they are welcome to .

- We ideally want this for all hash functions, however this will need to be done iteratively over time.

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

kevaundray:
# Related issue(s)

Resolves #313 

# Description

## Summary of changes

(If applicable.)

## Test additions / changes

(If applicable.)

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/acir/src/circuit/opcodes/black_box_function_call.rs
+++ b/acir/src/circuit/opcodes/black_box_function_call.rs
@@ -73,6 +73,10 @@ pub enum BlackBoxFuncCall {
     },
     Keccak256VariableLength {
         inputs: Vec<FunctionInput>,
+        /// This is the number of bytes to take
+        /// from the input. Note: if `var_message_size`
+        /// is more than the number of bytes in the input,
+        /// then an error is returned.
         var_message_size: FunctionInput,
         outputs: Vec<Witness>,
     },
--- a/acir/src/circuit/opcodes/black_box_function_call.rs
+++ b/acir/src/circuit/opcodes/black_box_function_call.rs
@@ -76,6 +76,11 @@ pub enum BlackBoxFuncCall {
         inputs: Vec<FunctionInput>,
         outputs: Vec<Witness>,
     },
+    Keccak256VariableLength {
+        inputs: Vec<FunctionInput>,
+        var_message_size: FunctionInput,
+        outputs: Vec<Witness>,
+    },
 }
 
 impl BlackBoxFuncCall {
@@ -139,6 +144,7 @@ impl BlackBoxFuncCall {
             BlackBoxFuncCall::EcdsaSecp256k1 { .. } => BlackBoxFunc::EcdsaSecp256k1,
             BlackBoxFuncCall::FixedBaseScalarMul { .. } => BlackBoxFunc::FixedBaseScalarMul,
             BlackBoxFuncCall::Keccak256 { .. } => BlackBoxFunc::Keccak256,
+            BlackBoxFuncCall::Keccak256VariableLength { .. } => BlackBoxFunc::Keccak256,
         }
     }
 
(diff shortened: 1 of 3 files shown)


acir/src/circuit/opcodes/black_box_function_call.rs
```diff
@@ -71,6 +71,11 @@ pub enum BlackBoxFuncCall {
         inputs: Vec<FunctionInput>,
         outputs: Vec<Witness>,
     },
+    Keccak256VariableLength {
+        inputs: Vec<FunctionInput>,
+        var_message_size: FunctionInput,
```
vezenovm, line 76: Could we specify what `var_message_size` represents? Is it the number of inputs to accept? Or is it the number of bytes of the full message (I know from pwg/hash.rs that it is the num bytes but this is unclear to me on its own)?

kevaundray: Good point -- I'll add a comment

kevaundray: Something else to add is a safety check for when `var_message_size` is more than the input bytes -- We could define the behaviour as taking all of the message of just returning an error. I'll return an error as its the safer option and users will need to constrain `var_message_size` appropriately.

vezenovm (approved):

TomAFrench:
Do we need to have two variants here? What are the benefits of this compared to just having the variable version and passing in `msg.len()` for the fixed size keccak?

kevaundray:
> Do we need to have two variants here? What are the benefits of this compared to just having the variable version and passing in `msg.len()` for the fixed size keccak?

They have different implementations on the backend, so if we put the msg.len() parameter on Keccak, then backends will need to implement KeccakVar which is harder to implement than Keccak. 

Theres also the fact that regular Keccak is more performant when `var_message_size` is a constant, I don't know how much more performant it is though

TomAFrench (approved):
Ok cool, this looks sensible if we're going down the two opcodes route.

Shall we add a todo to add variable variants of the other hash functions?

kevaundray:
> Ok cool, this looks sensible if we're going down the two opcodes route.
> 
> Shall we add a todo to add variable variants of the other hash functions?

Yep sure, made #320

Issue #326, opened 2023-05-30
chore: Remove Field < and <= support from Brillig


ludamad:
### Problem

See convo with Kev:
```
Essentially it was:
pub enum BinaryFieldOp {
    Add,
    Sub,
    Mul,
    Div,
    Cmp(Comparison),
}
Where we implement comparison for Field operations
In Noir we don't implement < and > for Field operations because technically fields are not ordered
```

### Happy Case

Just a single Equals opcode for BinaryFieldOp. Also refactor Int to be similar (collapse Comparison type)

### Would you like to submit a PR for this Issue?

Yes

ludamad:
Closed by https://github.com/noir-lang/acvm/pull/328

PR #328, opened 2023-05-30
chore: Brillig BinaryFieldOp simplification, rename resolve_foreign_call_input

### LINKED ISSUE #306: Consider renaming `resolve_foreign_call_input` in Brillig VM
### Problem

Comment regarding this is [here](https://github.com/noir-lang/acvm/pull/152#discussion_r1200937418)

This method does not seem to have anything to do with foreign calls in isolation, so a method stating what it is doing would be better.

### Proposed solution

- Add doc comment to function.
- Consider renaming it to something more appropriate

### Alternatives considered

_No response_

### Additional context

_No response_

### Submission Checklist

ludamad:
Resolves #306 #326

# Description

Various fixup changes. Includes a change to try and catch misuses of registers (by having an explicit uninitialized value)

## Summary of changes

Rename resolve_foreign_call_input for #306 
chore: Remove Field < and <= support from Brillig for #326 

N/A

## Test additions / changes

Some minor test simplification as field LTE was no longer available

# Checklist

# Additional context

(If applicable.)

### DIFF
--- a/brillig_vm/src/lib.rs
+++ b/brillig_vm/src/lib.rs
@@ -43,23 +43,25 @@ pub struct ForeignCallResult {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-// VM encapsulates the state of the Brillig VM during execution.
+/// VM encapsulates the state of the Brillig VM during execution.
 pub struct VM {
-    // Register storage
+    /// Register storage
     registers: Registers,
-    // Instruction pointer
+    /// Instruction pointer
     program_counter: usize,
-    // Tracks foreign calls
+    /// A counter maintained throughout a Brillig process that determines
+    /// whether the caller has resolved the results of a [foreign call][Opcode::ForeignCall].
     foreign_call_counter: usize,
-    // Holds foreign call results
+    /// Represents the outputs of all foreign calls during a Brillig process
+    /// List is appended onto by the caller upon reaching a [VMStatus::ForeignCallWait]
     foreign_call_results: Vec<ForeignCallResult>,
-    // Executable opcodes
+    /// Executable opcodes
     bytecode: Vec<Opcode>,
-    // Status of the VM
+    /// Status of the VM
     status: VMStatus,
-    // Memory of the VM
+    /// Memory of the VM
     memory: Vec<Value>,
-    // Call stack
+    /// Call stack
     call_stack: Vec<Value>,
 }
 
--- a/brillig_vm/src/lib.rs
+++ b/brillig_vm/src/lib.rs
@@ -10,8 +10,8 @@ mod opcodes;
 mod registers;
 mod value;
 
+pub use opcodes::Opcode;
 pub use opcodes::{BinaryFieldOp, BinaryIntOp, RegisterValueOrArray};
-pub use opcodes::{Comparison, Opcode};
 pub use registers::{RegisterIndex, Registers};
 use serde::{Deserialize, Serialize};
 pub use value::Typ;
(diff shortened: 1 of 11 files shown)


brillig_vm/src/lib.rs
```diff
@@ -471,6 +464,7 @@ mod tests {
 
     #[test]
     fn cmp_binary_ops() {
+        let bit_size = 32;
```
ludamad, line 467: Previously this was not correct if both elements are modulo before the result (which I assume the intent is - to first cast to N bit integers)

brillig_vm/src/lib.rs
```diff
@@ -300,14 +306,6 @@ impl VM {
     }
 }
 
-pub struct VMOutputState {
-    pub registers: Registers,
```
ludamad, line 304: Unused

brillig_vm/src/registers.rs
```diff
@@ -1,42 +1,23 @@
...
-/// Registers will store field element values during the
-/// duration of the execution of the bytecode.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Registers {
-    pub inner: Vec<Value>,
-}
-
-impl IntoIterator for Registers {
-    type Item = Value;
-    type IntoIter = RegistersIntoIterator;
```
ludamad, line 13: Unused. I don't think anyone needs to iterate register space?

kevaundray: Right, I think if it's not used internally then it's not needed

brillig_vm/src/registers.rs
```diff
@@ -1,42 +1,23 @@
...
-        RegistersIntoIterator { registers: self, index: 0 }
-    }
-}
-pub struct RegistersIntoIterator {
-    registers: Registers,
-    index: usize,
+    // Keep track of registers as an option to represent uninitialized registers
+    // These occur when we set a register value at a noncontiguous index, past currently defined registers
+    // This could just store 0's, but it would be
+    pub inner: Vec<Option<Value>>,
```
dbanks12, line 9: Unfinished comment? Also this is still a bit confusing to me.... Can't follow why you'd need/want this.

brillig_vm/src/registers.rs
vezenovm: A register iterator was used previously when setting the brillig outputs in the BrilligSolver, but that has changed now and we can remove this

brillig_vm/src/lib.rs
```diff
@@ -43,22 +43,28 @@ pub struct ForeignCallResult {
...
+    // Instruction pointer
     program_counter: usize,
-    /// A counter maintained throughout a Brillig process that determines
-    /// whether the caller has resolved the results of a [foreign call][Opcode::ForeignCall].
+    // Tracks foreign calls
     foreign_call_counter: usize,
-    /// Represents the outputs of all foreign calls during a Brillig process
-    /// List is appended onto by the caller upon reaching a [VMStatus::ForeignCallWait]
+    // Holds foreign call results
     foreign_call_results: Vec<ForeignCallResult>,
```
dbanks12, line 55: Are these removed comments no longer true? I think they are helpful.

dbanks12 (approved):
LGTM, couple of minor comments.

brillig_vm/src/lib.rs
vezenovm: I am curious about this removal as well

ludamad: Just an accident

Issue #300, opened 2023-05-22
Re-export black box functions for ACVM simulator to expose


TomAFrench:
### Problem

The ACVM simulator has an open issue to expose implementation of the black box function solvers: https://github.com/noir-lang/acvm-simulator/issues/5.

We want to enforce consistency rather than having a separate implementation in the simulator so ideally we'd reuse the solvers which exist within ACVM however these have a pretty non-ideal interface, e.g.:

```
sha256(
    initial_witness: &mut WitnessMap,
    inputs: &[FunctionInput],
    outputs: &[Witness],
)
```

### Proposed solution

To ensure consistency between ACVM and the simulator we should rework the underlying solvers within ACVM to be "pure" and then wrap these in a function which handles writing to the witness map.

### Submission Checklist

- [ ] Once I hit submit, I will assign this issue to the Project Board with the appropriate tags.

PR #359, opened 2023-06-10
chore: enforce `verify_prehashed` input lengths through types

TomAFrench:
# Description

## Problem\*

Related to #300 
## Summary\*

This PR sets out to enforce the proper lengths of inputs through the type system. This is useful for #300 as we'll be allowing user input here rather than this function only being called by the ACVM (which has stronger guarantees on passing correct length arrays.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/blackbox/ecdsa.rs
+++ b/acvm/src/pwg/blackbox/ecdsa.rs
@@ -87,8 +87,7 @@ mod ecdsa_secp256k1 {
         let message =
             b"ECDSA proves knowledge of a secret number in the context of a single message";
 
-        let digest: [u8; 32] =
-            Sha256::digest(message).try_into().expect("SHA256 digest should be 256 bits");
+        let digest = Sha256::digest(message);
 
         let signature: Signature = signing_key.sign(message);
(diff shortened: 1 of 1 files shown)


acvm/src/pwg/blackbox/ecdsa.rs
```diff
@@ -27,6 +27,17 @@ pub(crate) fn secp256k1_prehashed(
     hashed_message_inputs: &[FunctionInput],
     output: Witness,
 ) -> Result<OpcodeResolution, OpcodeResolutionError> {
+    let hashed_message: [u8; 32] =
+        to_u8_vec(initial_witness, hashed_message_inputs)?.try_into().map_err(|_| {
```
kevaundray, line 31: The hashed_message does not need to be 32 bytes. Barretenberg will crash if one supplies a hashed_message thats more than 32 bytes, but the algorithm itself is for a generic hash.

TomAFrench: Gotcha, I've relaxed this to accept a slice for the message hash then.

kevaundray (approved):

PR #363, opened 2023-06-12
feat: add method to generate the new `Brillig` opcode from `UnresolvedBrilligCall`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR aims to smooth out the devex of resolving a brillig foreign call by providing a helper function to generate an updated `Brillig` using the returned call result.

This allows us to use the snippet below

```rust
let result: ForeignCallResult = Oracle::do_brillig_foreign_call(unresolved_brillig_call.foreign_call_wait_info);
let new_brillig: Brillig = unresolved_brillig_call.resolve_foreign_call(result);
opcodes.push(Opcode::Brillig(new_brillig))
```

as opposed to having to extract the `Brillig` struct from the foreign call and manually push a new `ForeignCallResult` onto one of its fields.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acir/src/circuit/brillig.rs
+++ b/acir/src/circuit/brillig.rs
@@ -22,8 +22,10 @@ pub enum BrilligOutputs {
 pub struct Brillig {
     pub inputs: Vec<BrilligInputs>,
     pub outputs: Vec<BrilligOutputs>,
-    /// Results of oracles/functions external to brillig like a database read
+    /// Results of oracles/functions external to brillig like a database read.
+    // Each element of this vector corresponds to a single foreign call but may contain several values.
     pub foreign_call_results: Vec<ForeignCallResult>,
+    /// The Brillig VM bytecode to be executed by this ACIR opcode.
     pub bytecode: Vec<brillig_vm::Opcode>,
     /// Predicate of the Brillig execution - indicates if it should be skipped
     pub predicate: Option<Expression>,
--- a/acvm/src/pwg/mod.rs
+++ b/acvm/src/pwg/mod.rs
@@ -533,14 +533,17 @@ mod tests {
         assert_eq!(unsolved_opcodes.len(), 0, "brillig should have been removed");
         assert_eq!(unresolved_brillig_calls.len(), 1, "should have a brillig oracle request");
 
-        let UnresolvedBrilligCall { foreign_call_wait_info, mut brillig } =
-            unresolved_brillig_calls.remove(0);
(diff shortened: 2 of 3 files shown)


TomAFrench:
> Can you rewrite the Brillig oracle tests to use this new interface?

Do we have tests for how to resolve Brillig calls from outside the ACVM? I can only see the tests in `brillig_vm/src/lib.rs` which work on the Brillig VM directly (and so never create an `UnresolvedBrilligCall`.

vezenovm:
> Do we have tests for how to resolve Brillig calls from outside the ACVM?

They are the main pwg tests: https://github.com/noir-lang/acvm/blob/d5ba786e9bb6d965c55b393e189deda87c9faec3/acvm/src/pwg/mod.rs#L428

TomAFrench:
Although, come to think of it. By that logic maybe we should rename `resolve_foreign_call` to `resolve`.

vezenovm:
> although I think we should consider renaming the `foreign_call_wait_info` field of `UnresolvedBrilligCall` at some point

Yeah I agree. I think we can bikeshed this name in general. Perhaps just `wait_info` with the name type `ForeignCallWaitInfo`. Or we can bilkeshed when we fully get rid of the Oracle opcode

vezenovm (approved):

acir/src/circuit/brillig.rs
```diff
@@ -22,8 +22,10 @@ pub enum BrilligOutputs {
 pub struct Brillig {
     pub inputs: Vec<BrilligInputs>,
     pub outputs: Vec<BrilligOutputs>,
-    /// Results of oracles/functions external to brillig like a database read
+    /// Results of oracles/functions external to brillig like a database read.
+    // Each element of this vector corresponds to a single foreign call but may contain several values.
```
vezenovm, line 26: nit are we missing `/` here to make it a doc comment?

TomAFrench: This was intentional as I'm not sure it's something that needs to be known by a user but as I was piecing together how brillig foreign calls worked it wasn't immediately obvious. As it's only really needed by people working inside the ACVM (as we should abstract around this) then I think a normal comment is fine.

vezenovm: Ok sounds good

TomAFrench:
> Or we can bilkeshed when we fully get rid of the Oracle opcode

Yeah, I think we can make this decision later as we'll "own" most of the broken code but doing it at the same time as a breaking change of oracle code makes sense.

PR #365, opened 2023-06-12
fix(brillig): Explicitly wrap on arithmetic operations

### LINKED ISSUE #360: Explicitly define behaviour on overflow/underflow in Brillig operations
### Problem

On release mode, [the following operations](https://github.com/noir-lang/acvm/blob/4eefda01e7b371035314f77631df4687608b4782/brillig_vm/src/opcodes.rs#L167-L169) will overflow or underflow and not warn the user. While in debug mode, they will panic since we have not written exactly what behaviour we expect.

### Happy Case

Replace these with the explicit methods specifying what the behaviour should be.

For examples, see the documentation on u128 [here](https://doc.rust-lang.org/std/primitive.u128.html#method.wrapping_add)

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

sirasistant:
# Description

## Problem\*

Resolves #360

## Summary\*

This PR sets out to

### Example

Before:

```

```

After:

```

```

## Additional Context

# PR Checklist\*

### DIFF
--- a/brillig_vm/src/opcodes.rs
+++ b/brillig_vm/src/opcodes.rs
@@ -327,6 +327,7 @@ mod tests {
             TestParams {
                 a: to_negative(2, bit_size),
                 b: 7,
+                // negative 14 wraps to a 2
                 result: to_negative(14, bit_size),
             },
         ];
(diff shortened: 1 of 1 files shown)


ludamad:
I'll let @kevaundray look but in the future should I just merge these if I'm confident? Or would you like final say?

kevaundray:
> I'll let @kevaundray look but in the future should I just merge these if I'm confident? Or would you like final say?

Yeah its fine to merge these Brillig based PRs tentatively

PR #367, opened 2023-06-12
feat(brilig)!: Multiple foreign call inputs 

### LINKED ISSUE #366: Expand Brillig Foreign Call to accept multiple inputs/outputs
### Problem

Currently the Brillig ForeignCall opcode only accepts a singualr `RegisterValueOrArray` for both its inputs and its outputs. This is too restrictive as this forces the compiler generating Brillig to flatten all of its inputs/outputs to a singular array if it has multiple inputs.

### Happy Case

We can generate Brillig code for a foreign call that has multiple inputs/outputs.

### Alternatives Considered

It is still possible to have multiple inputs/outputs in Brillig by flattening all inputs/outputs into a single array. However, this leads to needing the external structure of the IR which is generating Brillig code in order to interpret the structure being passed into a foreign call. We want to be able generate and interpret Brillig without any external IR info.

### Additional Context

_No response_

No

### Support Needs

_No response_

vezenovm:
# Description

Extends the `ForeignCall` opcode to accept multiple `RegisterValueOrArray` inputs and destinations slots. 

## Problem\*

Resolves #366 

## Summary\*

The PR changes the interface for interacting with foreign calls where a foreign call can accept multiple inputs and outputs.  Previously in order to accept or return multiple values from a foreign call would require flattening by the one generating Brillig. This also makes Brillig much harder to interpret on its own as it can mean different things depending on the IR interface that generated Brillig. Being explicit about many inputs to/from a foreign call enables our bytecode to be more clear. 

### Example

Before:

```

```

After:

```

```

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/brillig.rs
+++ b/acvm/src/pwg/brillig.rs
@@ -142,5 +142,5 @@ pub struct ForeignCallWaitInfo {
     /// An identifier interpreted by the caller process
     pub function: String,
     /// Resolved inputs to a foreign call computed in the previous steps of a Brillig VM process
-    pub inputs: Vec<Value>,
+    pub inputs: Vec<Vec<Value>>,
 }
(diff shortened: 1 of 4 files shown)


ludamad (approved):
LGTM pending @kevaundray's approval of the move (made sense to me for an overall simplification POV)

PR #369, opened 2023-06-13
chore(brillig_vm): Expand memory on ForeignCall outputs

vezenovm:
# Description

## Problem\*

Resolves 

## Summary\*

If someone decides to make a foreign call that returns an array the Brillig compiler will be required to write extra opcodes (Const and Store) in order to allocate memory for the foreign call result if they do not want to overwrite previous memory locations. We currently expand when we reach memory limits on store and we should match this functionality for ForeignCall. 

### Example

I wrote a test showing this allocation for ForeignCall array outputs

Before:

```

```

After:

```

```

## Additional Context

# PR Checklist\*


ludamad (approved):
LGTM

PR #370, opened 2023-06-13
feat(brillig): Allow dynamic-size foreign calls

ludamad:
# Description

## Problem\*

Incremental progress towards https://github.com/noir-lang/noir/issues/1556

Takes this milestone "Add support for slices such that in Brillig we can call an oracle and have it return a slice with length not known and enumerate over that slice" and does the part necessary for such oracles in brillig vm

As an intro, start with the comment in brillig_vm/src/opcodes.rs

## Additional Context

# PR Checklist\*

- [*] I have tested the changes locally.
- [*] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.


brillig_vm/src/lib.rs
```diff
@@ -923,6 +943,72 @@ mod tests {
...
+        let r_input_pointer = RegisterIndex::from(0);
+        let r_input_size = RegisterIndex::from(1);
+        // We need to pass a location of appropriate size
+        let r_output_pointer = RegisterIndex::from(2);
+        let r_output_size = RegisterIndex::from(3);
+
+        // Our first string to use the identity function with
+        let input_string =
+            vec![Value::from(1u128), Value::from(2u128), Value::from(3u128), Value::from(4u128)];
+        // Double the string (concate it with itself)
```
kevaundray: typo: concate

ludamad:
Very weird thing - when I closed my unconventionally-titled PR (had a DONT MERGE at start), this one started to report that failure

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
...
 
+/// Lays out various ways we might interpret memory when passing to a foreign call external from Brillig.
+///
+/// While we are usually agnostic to how memory is passed within Brillig,
+/// this needs to be encoded somehow when dealing with an external system.
+/// For simplicity, the extra type information is given right in the ForeignCall instructions.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Copy)]
-pub enum RegisterValueOrArray {
+pub enum RegisterOrMemory {
+    /// An immediate value passed to or from an external call
```
vezenovm, line 14: What does "immediate value" mean vs just a "single value" ?

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
 
 pub type Label = usize;
 
+/// Lays out various ways we might interpret memory when passing to a foreign call external from Brillig.
```
vezenovm, line 7:
```suggestion
/// Lays out various ways an external foreign call's input and output data may be interpreted inside Brillig.
/// This data can either be an individual register value or memory.
```

vezenovm: Just wanted to highlight more clearly what is meant to pass to a foreign call. Does that mean the foreign call inputs or outputs? The enum represents both inputs/outputs so wanted to highlight that. And then inside the enumeration we go over what happens in both the input and output case. Feel free to accept the above suggestion or rewrite how you like

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
...
+///
+/// While we are usually agnostic to how memory is passed within Brillig,
+/// this needs to be encoded somehow when dealing with an external system.
+/// For simplicity, the extra type information is given right in the ForeignCall instructions.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Copy)]
-pub enum RegisterValueOrArray {
+pub enum RegisterOrMemory {
+    /// An immediate value passed to or from an external call
+    /// For a foreign call input, this is read directly from the register.
```
vezenovm, line 15:
```suggestion
    /// For a foreign call input, the value is read directly from the register.
```

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
...
+    /// For a foreign call output, this is written directly to the register.
```
vezenovm, line 16:
```suggestion
    /// For a foreign call output, the value is written directly to the register.
```

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
...
     RegisterIndex(RegisterIndex),
+    /// A fix-sized array passed starting from a Brillig register memory location.
+    /// In the case of a foreign call input, this is read from this Brillig memory location + usize more cells.
+    /// In the case of a foreign call output, this is written to this Brillig memory location with the usize being here just as a sanity check for the size write.
```
vezenovm, line 20:
```suggestion
    /// In the case of a foreign call input, the array is read from this Brillig memory location + usize more cells.
    /// In the case of a foreign call output, the array is written to this Brillig memory location with the usize being here just as a sanity check for the size write.
```

brillig_vm/src/opcodes.rs
```diff
@@ -4,10 +4,25 @@ use serde::{Deserialize, Serialize};
...
     HeapArray(RegisterIndex, usize),
+    /// A register-sized vector passed starting from a Brillig register memory location and with a register-held size
+    /// In the case of a foreign call input, this is read from this Brillig memory location + as many cells as the 2nd register indicates.
+    /// In the case of a foreign call output, this is written to this Brillig memory location with the usize being here just as a sanity check for the size write.
```
vezenovm, line 24:
```suggestion
    /// In the case of a foreign call input, the vector is read from this Brillig memory location + as many cells as the 2nd register indicates.
    /// In the case of a foreign call output, the vector is written to this Brillig memory location with the usize being here just as a sanity check for the size write.
```

brillig_vm/src/opcodes.rs
vezenovm: This comment and the ones below on the enum variants are just nits that I thought made things a bit clearer

brillig_vm/src/lib.rs
```diff
@@ -205,32 +205,46 @@ impl VM {
 
                 for (destination, values) in destinations.iter().zip(values) {
                     match destination {
-                        RegisterValueOrArray::RegisterIndex(index) => {
+                        RegisterOrMemory::RegisterIndex(value_index) => {
                             assert_eq!(
                                 values.len(),
                                 1,
-                                "Function result size does not match brillig bytecode"
+                                "Function result size does not match brillig (expected 1 result)"
```
vezenovm, line 212: This is a nit but we should settle on when we refer to just "Brillig" if we mean the VM or the bytecode. I lean towards that Brillig is the VM interface. I think here "brillig bytecode" is a bit clearer as it says the we do not match the specification in the ForeignCall opcode. Same for the other asserts in this loop.

brillig_vm/src/lib.rs
```diff
@@ -923,6 +943,71 @@ mod tests {
...
+        vm.foreign_call_results.push(ForeignCallResult { values: vec![output_string.clone()] });
+
+        // Resume VM
+        brillig_execute(&mut vm);
+
+        // Check that VM finished once resumed
+        assert_eq!(vm.status, VMStatus::Finished);
+
+        // Check result in memory
+        let result_values = vm.memory[0..output_string.len()].to_vec();
```
vezenovm, line 1004: We cloned the input string twice exactly so 0..output_string.len() would trivially match the output string. If we reversed the output_string after constructing it (before resolving the foreign call) we would get a failure as this is reading the input string and first half of the output string. Switching this to `input_string.len()..(input_string.len() + output_string.len())` makes the test more accurate

vezenovm, line 1004:
```suggestion
        let result_values = vm.memory[input_string.len()..(input_string.len() + output_string.len())].to_vec();
```

vezenovm:
Mostly nit comments but this looks solid

brillig_vm/src/lib.rs
vezenovm: I ended up pushing a commit directly here that changed this test to reverse the concatenation (https://github.com/noir-lang/acvm/pull/370/commits/dd05610dcdab4f1f383e42f772c829ab22f2293d) to make the test more thorough

brillig_vm/src/opcodes.rs
ludamad: Immediate because it's not in memory

ludamad:
@vezenovm I merged your work into here, this could use another pass

brillig_vm/src/lib.rs
```diff
@@ -203,36 +209,66 @@ impl VM {
...
+                        RegisterOrMemory::HeapVector(pointer_index, size_index) => {
+                            match output {
+                                ForeignCallOutput::Array(values) => {
+                                    // Convert the size pointer to a usize
+                                    let size = self.registers.get(*size_index).to_usize();
+                                    assert_eq!(
+                                        values.len(),
+                                        size,
+                                        "Function result size does not match brillig bytecode size register"
+                                    );
```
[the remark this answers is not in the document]
TomAFrench: This isn't a blocker for this PR as we already have an assert here though.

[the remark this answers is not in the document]
ludamad: It's a good call, we can merge that first and ill update

vezenovm (approved):
@ludamad I resolved the merge conflicts w/ master after switching the way handle failure's for invalid foreign call results. I'm going to approve and you take a pass on my latest commits this PR looks good to me.

PR #374, opened 2023-06-14
chore(brillig_vm)!: Switch nested vector in ForeignCallResult to enum ForeignCallOutput

vezenovm:
# Description

## Problem\*

Cleanup ForeignCallResult to use an enum ForeignCallOutput rather than a nested vector. This makes the structure of each individual output clearer to developers and no longer mixes a single value w/ an array of size 1.

Resolves 

## Additional Context

# PR Checklist\*


ludamad:
I suppose it was getting confusing with the nesting, LGTM

PR #375, opened 2023-06-14
chore: set Brillig VM state to `Failure` rather than panicking on invalid foreign call response

TomAFrench:
# Description

## Problem\*

Returning an incorrectly sized foreign call response to the Brillig VM will currently cause a panic. This is an issue for situations where the Brillig VM is running inside wasm as this doesn't handle panics well. As this is a case of a user providing invalid input, we should return an error rather than panicking.

Resolves https://github.com/noir-lang/acvm/pull/370#discussion_r1229659602

## Summary\*

We now set the `VMStatus` to `Failure` in this situation. I've had to delay this until after the `for-loop` due to the borrow-checker so I'd be open to any other solutions which avoid the need for the mutable bool + break.

## Additional Context

# PR Checklist\*

### DIFF
--- a/brillig_vm/src/lib.rs
+++ b/brillig_vm/src/lib.rs
@@ -203,22 +203,23 @@ impl VM {
                 let ForeignCallResult { values } =
                     &self.foreign_call_results[self.foreign_call_counter];
 
+                let mut invalid_foreign_call_result = false;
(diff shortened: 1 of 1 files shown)


ludamad (approved):
Looks good to me

PR #378, opened 2023-06-15
feat(acvm)!: Replace `PartialWitnessGenerator` trait with `BlackBoxFunctionSolver`

### LINKED ISSUE #377: Replace `PartialWitnessGenerator` trait with `BlackBoxFunctionSolver` trait.
### Problem

As we've restricted the role of the `PartialWitnessGenerator` trait in backends, it no longer makes sense to give it full read/write access to the `WitnessMap` and it should only the core of the black box functions which it's providing with the rest of the logic living in ACVM.

The current state of this trait requires duplicated logic which will be just copy-pasted across backends (see [acvm-simulator](https://github.com/noir-lang/acvm-simulator/blob/a3fcf9601f87953885ac8653ad8a2ab2116518c7/src/execute.rs#L30-L140) and [acvm-backend-barretenberg](https://github.com/noir-lang/acvm-backend-barretenberg/blob/9013d1f88377618bee42bfc8cc95f36e5cab4c17/src/acvm_interop/pwg.rs#L12-L141)). Having many duplicates of this logic which has full write access to the `WitnessMap` can then lead to correctness issues.

### Happy Case

We should bring all of the logic for reading/writing from the witness map into ACVM. We'd write solvers for these three black box functions similarly to how we do for the rust implemented black box functions which handle reading/writing to the witness map.

The `PartialWitnessGenerator` trait would then become:
```rust
// Name is subject to feedback but I don't think `PartialWitnessGenerator` fits what this is anymore.
pub trait BlackBoxFunctionSolver {
    fn schnorr_verify(
        &self,
        public_key_x: &FieldElement,
        public_key_y: &FieldElement,
        signature: (&FieldElement, &FieldElement),
        message: Vec<u8>,
    ) -> Result<bool, OpcodeResolutionError>;
    fn pedersen(
        &self,
        inputs: &[FieldElement],
        domain_separator: u32,
    ) -> Result<(FieldElement, FieldElement), OpcodeResolutionError>;
    fn fixed_base_scalar_mul(
        &self,
        input: &FieldElement,
    ) -> Result<(FieldElement, FieldElement), OpcodeResolutionError>;
}
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

Yes

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #377 

## Summary\*

This PR removes the backend's access to the `WitnessMap` and simplifies it to just returning outputs for black box functions given a set of inputs.

Draft pending some questions about expected behaviour of schnorr opcode which affects the interface.

## Additional Context

# PR Checklist\*


acvm/src/pwg/blackbox/schnorr_verify.rs
```diff
@@ -0,0 +1,81 @@
...
+            )
+        })?;
+        let sig_i = witness_to_value(initial_witness, _sig_i.witness)?;
+        *sig = *sig_i.to_be_bytes().last().ok_or_else(|| {
+            OpcodeResolutionError::BlackBoxFunctionFailed(
+                BlackBoxFunc::SchnorrVerify,
+                "could not get last bytes".into(),
+            )
+        })?;
+    }
```
TomAFrench, line 60: @kevaundray My question on slack is basically around these lines. We restrict the public key to be a pair of elements of the underlying curve field so it stands to reason that the signature would be the same (and so can be represented as a pair of field elements) but we previously allowed a full u256 range.

acvm/src/pwg/blackbox/schnorr_verify.rs
TomAFrench, line 1: We'll probably want to reinstate the `signature` module and have this and ecdsa live together inside it.

acvm/src/pwg/blackbox/schnorr_verify.rs
TomAFrench: Closing as this is now a concern of #386

TomAFrench:
This PR should be gtg now.

kobyhallx (approved):
LGTM

Issue #379, opened 2023-06-15
Brillig: Make empty call stack return == stop


ludamad:
### Aim

This makes code linking more elegant, and allows functions to be used simply as entry points

### Expected Behavior

-

### Bug

-

### To Reproduce

1.
2.
3.
4.

### Would you like to submit a PR for this Issue?

No

PR #381, opened 2023-06-16
feat(brillig): Allow returning with an empty call stack

ludamad:
See https://github.com/noir-lang/acvm/issues/379

### DIFF
--- a/brillig_vm/src/lib.rs
+++ b/brillig_vm/src/lib.rs
@@ -181,7 +181,7 @@ impl VM {
                 if let Some(register) = self.call_stack.pop() {
                     self.set_program_counter(register.to_usize())
                 } else {
-                    self.fail("return opcode hit, but callstack already empty".to_string())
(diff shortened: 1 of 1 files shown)


ludamad:
Actually questioning this now. We still need entry point wrappers around functions if we start to do any call-convention-transforming optimizations, and for mundane setup that all the functions need like just initializing static data

ludamad:
@kevaundray thoughts on this?

ludamad:
@kevaundray ping

kevaundray:
If theres no strong conviction to add this then I'd opt to not merge and leave the issue open, so that we can discuss this better when Brillig is in a more stable state. I'd lean towards the status quo since it forces us to be intentional about the way we handle call stacks

TomAFrench:
Based on the above, I'm going to close this PR. We can reopen it after the discussions have been had.

PR #368, opened 2023-06-12
feat(acir)!: Remove `Oracle` opcode

TomAFrench:
# Description

## Problem\*

Resolves #298 

## Summary\*

This PR removes the `Oracle` opcode as it has been superceded by Brillig foreign calls.

## Additional Context

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #384, opened 2023-06-16
feat(acvm)!: Encapsulate internal state of ACVM within a struct

### LINKED ISSUE #383: Refactor ACVM solver to allow tracking internal state
### Problem

Much of the state transitions in the ACVM are unnecessarily exposed or even outsourced to the caller due to the fact that the ACVM has no internal state. This causes issues such as described in #382. For example:

- When the ACVM encounters a foreign call, it aborts execution while returning the foreign calls along with a list of unsolved opcodes. The caller is then expected to manipulate those foreign call's directly, create new opcodes to represent their resolution, insert these handcrafted opcodes into the list of opcodes to execute and then restart the ACVM with this new set of opcodes.
- We require the caller to construct a `Blocks` instance and pass in a mutable reference to `acvm::pwg:solve`. They must then persist this across multiple invocations of `acvm::pwg:solve` as the ACVM cannot maintain track over its partial execution of `Block` opcodes itself.

### Happy Case

Rather than having a stateless function we can represent the ACVM as an `ACVM` struct as we do for the Brillig VM.

```rust
struct ACVM {
  backend: PartialWitnessGenerator
  witness_map: WitnessMap,
  blocks: Blocks
  opcodes: Vec<Opcode> 
}

impl ACVM {
  fn solve() -> ACVMStatus {
    // Do stuff
  }
}
```

This struct would encapsulate any internal state of the ACVM.

### Alternatives Considered

_No response_

### Additional Context

_No response_

Yes

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #383 

## Summary\*

This PR creates an struct to track the internal state of the ACVM execution.

## Additional Context

# PR Checklist\*


TomAFrench:
This can be simplified significantly with #368 and #385. We will then also be able to remove all cloning of opcodes.

acvm/src/pwg/mod.rs
```diff
@@ -85,84 +82,109 @@ pub enum OpcodeResolutionError {
...
+                    }
+                    Err(OpcodeResolutionError::OpcodeNotSolvable(_)) => {
+                        unreachable!("ICE - Result should have been converted to GateResolution")
+                    }
+                    Err(err) => return Err(err),
                 }
-                Err(err) => return Err(err),
+            }
+
+            std::mem::swap(&mut self.opcodes, &mut unresolved_opcodes);
```
TomAFrench, line 170: I'd like to have a little bit extra scrutiny on this (with an extra test case most likely). This looks correct but we want to ensure that the internal state is updated correctly when performing an foreign call.

acvm/src/pwg/mod.rs
```diff
@@ -35,11 +35,8 @@ pub enum PartialWitnessGeneratorStatus {
     /// to retrieve information from outside of the ACVM.
     /// The result of the foreign call is inserted into the `Brillig` opcode which made the call using [`UnresolvedBrilligCall::resolve`].
     ///
-    /// Once this is done, the `PartialWitnessGenerator` can be restarted to solve the new set of opcodes.
-    RequiresForeignCall {
-        unsolved_opcodes: Vec<Opcode>,
-        unresolved_brillig_calls: Vec<UnresolvedBrilligCall>,
-    },
+    /// Once this is done, the `PartialWitnessGenerator` can be restarted to solve the remaining opcodes.
```
TomAFrench, line 38: Doc comment needs to be updated.

acvm/src/pwg/mod.rs
```diff
@@ -86,84 +83,136 @@ pub enum OpcodeResolutionError {
...
+
+    /// Resolves a pending foreign call using a result calculated outside of the ACVM.
+    pub fn resolve_pending_foreign_call(&mut self, foreign_call_result: ForeignCallResult) {
+        // Remove the first foreign call and inject the result to create a new opcode.
+        let foreign_call = self.pending_foreign_calls.remove(0);
+        let resolved_brillig = foreign_call.resolve(foreign_call_result);
+
+        // Mark this opcode to be executed next.
+        self.opcodes.insert(0, Opcode::Brillig(resolved_brillig));
+    }
```
TomAFrench, line 134: May want to adjust this to a format which allows parallel resolution of these.

acvm/src/pwg/mod.rs
```diff
@@ -86,84 +83,141 @@ pub enum OpcodeResolutionError {
...
+            opcodes,
+            witness_map: initial_witness,
+            pending_foreign_calls: Vec::new(),
+        }
+    }
+
+    // necessary to fix integration tests. Should replace with a better VM status model.
+    pub fn remaining_opcodes_len(&self) -> usize {
+        self.opcodes.len()
+    }
```
TomAFrench, line 114: This should be removed before merging.

kevaundray:
Is this aiming to encapsulate ACVM or just the PWG?

TomAFrench:
> ACVM involves both a solver and a backend. Where the backend does proving and verifying

Is that the correct way to think about it? I mean, yes if we're talking about "ACVM the repository" then this naming is suboptimal but if we're thinking about "ACVM the virtual machine" then the struct naming is accurate.

If we were talking about a zkEVM, I don't think we'd include the prover/verifier in the definition of the zkEVM but accompanying components. The zkEVM is just the circuits which take initial state and generate the output state which is similar to what the ACVM struct is doing here.

kevaundray:
> Is that the correct way to think about it? I mean, yes if we're talking about "ACVM the repository" then this naming is suboptimal but if we're thinking about "ACVM the virtual machine" then the struct naming is accurate.

I've always thought of it as ACVM being instantiated with a backend but I can see where you are coming from. If thats the case, then wouldn't proving and verifying being in this repository be an abstraction leak?

TomAFrench:
> If thats the case, then wouldn't proving and verifying being in this repository be an abstraction leak?

Potentially, yes. The `SmartContract` and `ProofSystemCompiler` traits depend on ACIR rather than the ACVM itself. These interfaces could be defined elsewhere as some for of "surrounding tooling" for the ACVM.

acvm/src/pwg/mod.rs
TomAFrench: I'm going to hold off of this for the time being as we may end up removing the ability to get into the situation where multiple parallel foreign calls is possible.

TomAFrench:
I don't think this will be the final state of the `ACVM` struct but its in a mergeable condition.

ludamad (approved):

PR #386, opened 2023-06-16
feat(acir)!: Use fixed length data structures in black box function inputs/outputs where possible.

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR modifies any opcodes which have a known number of inputs/outputs to hold a tuple of these values rather than a vector.

## Additional Context

# PR Checklist\*


TomAFrench:
We could also do this with the pubkey + signature for ECDSA but this causes issues with serialisation.

acvm/tests/solver.rs
```diff
@@ -49,7 +49,7 @@ impl PartialWitnessGenerator for StubbedPwg {
         &self,
         _initial_witness: &mut WitnessMap,
         _input: &FunctionInput,
-        _outputs: &[Witness],
+        _outputs: &(Witness, Witness),
```
kevaundray, line 52: You shouldn't need the `&` here since Witness is Copyable

acvm/tests/solver.rs
```diff
@@ -28,7 +28,7 @@ impl PartialWitnessGenerator for StubbedPwg {
         _initial_witness: &mut WitnessMap,
         _public_key_x: &FunctionInput,
         _public_key_y: &FunctionInput,
-        _signature: &[FunctionInput],
+        _signature: &(FunctionInput, FunctionInput),
```
kevaundray, line 31: same re FunctionInput being Copyable

acvm/src/lib.rs
```diff
@@ -74,24 +74,24 @@ pub trait PartialWitnessGenerator {
     fn schnorr_verify(
         &self,
         initial_witness: &mut WitnessMap,
-        public_key_x: &FunctionInput,
-        public_key_y: &FunctionInput,
-        signature: &[FunctionInput],
+        public_key_x: FunctionInput,
+        public_key_y: FunctionInput,
+        signature: (FunctionInput, FunctionInput),
```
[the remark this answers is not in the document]
kevaundray: If you want, I'd prefer `sig_s` and `sig_e` just because you can provide a bit more context versus the tuple where its not clear what the first item and second item are. It also would be harder to mess up named fields version structs.

This can ofcourse be aleviated with proper documentation, so no strong incling if you prefer the other

kevaundray (approved):
This looks good to me, once you've made a decision on naming consistency, feel free to ping me again!

TomAFrench:
I'm silencing the clippy warning as it will be addressed in #378

PR #390, opened 2023-06-19
chore: make tests for `acvm::pwg::solve` integration tests

TomAFrench:
# Description

## Problem\*

I noticed that we weren't exposing `ForeignCallWaitInfo` which made it harder to implement brillig call support in ACVM simulator.

## Summary\*

`acvm::pwg::solve` is the `acvm` crates' main public export however we don't test its usability from the perspective of an external user. We want to ensure that all the relevant types are public but this is not enforced by the current unit tests.

I've then moved them to be integration tests. These tests are integration tests anyway so it makes sense to move them here.

## Additional Context

# PR Checklist\*


kobyhallx (approved):
👍

PR #401, opened 2023-06-22
feat(brillig): implemented first blackbox functions

sirasistant:
# Description
Implements the blackbox opcodes and the first black box operations.

Here is a reference implementation of this PR in noir=>brillig https://github.com/noir-lang/noir/pull/1788

Also extracted HeapVector and HeapArray as structs same as RegisterIndex because they are handy for passing data to blackbox operations. 

Hashes are implemented with HeapVectors so they are as flexible as possible, since HeapArrays can be casted to HeapVector during compilation by just inssuing a Const opcode.
## Additional Context

# PR Checklist\*


brillig_vm/src/black_box.rs
```diff
@@ -0,0 +1,211 @@
...
+    for input in inputs {
+        let field_bytes = input.to_field().to_be_bytes();
+        let byte = field_bytes.last().unwrap();
+        result.push(*byte);
+    }
+    result
+}
+
+/// Does a generic hash of the inputs storing the resulting 32 bytes as items in the output array.
+fn generic_hash_256<D: Digest>(
```
kevaundray, line 121: We are delaying conversations around whether we should have this file share code with the blackbox function file in pwg, since @TomAFrench has been doing a lot of refactoring there, his input would be valuable here

brillig_vm/src/black_box.rs
```diff
@@ -0,0 +1,211 @@
...
+        hashed_msg: HeapVector,
+        public_key_x: HeapArray,
+        public_key_y: HeapArray,
+        signature: HeapArray,
+        result: RegisterIndex,
+    },
+}
+
+impl BlackBoxOp {
+    pub(crate) fn evaluate(&self, registers: &mut Registers, memory: &mut Memory) {
```
kevaundray, line 47: Can we add a test for one of these?

brillig_vm/Cargo.toml
```diff
@@ -13,6 +13,16 @@ repository.workspace = true
...
+blake2 = "0.10.6"
+sha2 = "0.10.6"
+sha3 = "0.10.6"
+k256 = { version = "0.11.0", features = [
+    "ecdsa",
+    "ecdsa-core",
+    "sha256",
+    "digest",
+    "arithmetic",
+] }
```
TomAFrench, line 25: We should make these workspace dependencies to ensure that Brillig + ACVM match.

brillig_vm/src/opcodes.rs
```diff
@@ -1,9 +1,23 @@
-use crate::{RegisterIndex, Value};
+use crate::{black_box::BlackBoxOp, RegisterIndex, Value};
 use acir_field::FieldElement;
 use serde::{Deserialize, Serialize};
 
 pub type Label = usize;
 
+/// A fix-sized array passed starting from a Brillig register memory location.
```
TomAFrench, line 7:
```suggestion
/// A fixed-sized array starting from a Brillig register memory location.
```

brillig_vm/src/black_box.rs
```diff
@@ -0,0 +1,211 @@
...
+        result.push(*byte);
+    }
+    result
+}
+
+/// Does a generic hash of the inputs storing the resulting 32 bytes as items in the output array.
+fn generic_hash_256<D: Digest>(
+    message: &HeapVector,
+    output: &HeapArray,
+    registers: &mut Registers,
```
TomAFrench, line 124:
```suggestion
    registers: &Registers,
```

We're not mutating any of the registers here.

brillig_vm/src/black_box.rs
sirasistant: That's true, the main point of reuse is the verify_secp256k1_ecdsa_signature but it's true that it's only one function. Maybe this gets more critical when we implement other BB functions on the ACVM side, such as pedersen, scalar mul, etc

kevaundray (approved):

PR #406, opened 2023-06-30
feat: Derive `Copy` for `Language`

phated:
# Description

## Problem\*

Resolves 

## Summary\*

This adds `Copy` to the languages. There's no reason we should need to pass references for `Language` because it can derive copy.

## Additional Context

This PR is applied to the 0.15.1 release so I can us it against noir but it should apply cleanly.

# PR Checklist\*


phated:
I just used `.clone()` in my Noir PR, so I'll update this to latest.

TomAFrench (approved):

PR #407, opened 2023-07-02
feat(stdlib): Add fallback implementation of `SHA256` black box function

Ethan-000:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*


stdlib/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![warn(unused_crate_dependencies)]
 #![warn(unreachable_pub)]
 
+pub mod blackbox_fallbacks;
 pub mod fallback;
```
TomAFrench, line 5: `mod fallback` already contains the fallback implementations of black box function. I'm not sure why we need another module here.

stdlib/src/blackbox_fallbacks/logic_fallbacks.rs
```diff
@@ -7,7 +7,7 @@ use acir::{
 
 // Perform bit decomposition on the provided expression
 #[deprecated(note = "use bit_decomposition function instead")]
-pub fn split(
+fn _split(
```
TomAFrench, line 10: I think we can remove this entirely tbh.

stdlib/src/blackbox_fallbacks/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub(crate) fn round_to_nearest_byte(num_bits: u32) -> u32 {
+    round_to_nearest_mul_8(num_bits) / 8
+}
+
+// TODO: Maybe this can be merged with `bit_decomposition`
+pub(crate) fn byte_decomposition(
+    gate: Expression,
+    num_bytes: u32,
+    mut num_witness: u32,
```
TomAFrench, line 31:
```suggestion
    num_witness: mut& u32,
```

Thoughts on passing a mutable reference here? We've got a _lot_ of extra return values everywhere due to needing to update `num_witness`.

stdlib/src/blackbox_fallbacks/logic_fallbacks.rs
```diff
@@ -5,19 +5,6 @@ use acir::{
     native_types::{Expression, Witness},
 };
 
-// Perform bit decomposition on the provided expression
-#[deprecated(note = "use bit_decomposition function instead")]
```
kevaundray, line 9: We should double check, this might be a breaking change, since it was marked as pub

stdlib/src/blackbox_fallbacks/uint32.rs
```diff
@@ -0,0 +1,620 @@
...
+        opcodes::{BlackBoxFuncCall, FunctionInput},
+        Opcode,
+    },
+    native_types::{Expression, Witness},
+    FieldElement,
+};
+
+/// UInt32 contains a witness that points to a field element that represents a u32 integer
+/// It has a inner field of type [Witness] that points to the field element and width = 32
+// TODO: This can be generalized to u8, u64 and others if needed.
```
kevaundray, line 16: For TODOs can you create an issue?

stdlib/src/blackbox_fallbacks/uint32.rs
```diff
@@ -0,0 +1,620 @@
...
+#[derive(Copy, Clone, Debug)]
+pub struct UInt32 {
+    pub inner: Witness,
```
kevaundray, line 19: Perhaps we can create a method which is only available for tests:


```
#[cfg(test)]
pub fn get_witness(&self) -> Witness {
   self.inner
}
```

Then the inner can be private

stdlib/src/blackbox_fallbacks/uint32.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+impl UInt32 {
+    /// Initialize A new [UInt32] type with a [Witness]
+    pub fn new(witness: Witness) -> Self {
+        UInt32 { inner: witness, width: 32 }
+    }
+
+    /// Load a [u128] constant into the circuit
+    // TODO: This is currently a u128 instead of a u32 cus
```
kevaundray, line 37: - create issue for TODO
- nit: cus -> because

stdlib/src/blackbox_fallbacks/uint32.rs
```diff
@@ -0,0 +1,620 @@
...
+            bytecode: vec![brillig_vm::Opcode::Stop],
+            predicate: None,
+        });
+        new_gates.push(brillig_opcode);
+        let num_witness = variables.finalize();
+
+        (UInt32::new(new_witness), new_gates, num_witness)
+    }
+
+    /// load a [UInt32] from four [Witness]es each representing a [u8]
```
kevaundray, line 64:
```suggestion
    /// Load a [UInt32] from four [Witness]es each representing a [u8]
```

stdlib/src/blackbox_fallbacks/uint32.rs
```diff
@@ -0,0 +1,620 @@
...
+            UInt32::load_constant(2_u128.pow(self.width), num_witness);
+        new_gates.extend(extra_gates);
+        let (_, sub_mod, extra_gates, num_witness) =
+            UInt32::euclidean_division(&UInt32::new(new_witness), &two_pow_width, num_witness);
+        new_gates.extend(extra_gates);
+
+        (sub_mod, new_gates, num_witness)
+    }
+
+    /// Caculate and constrain `self` - `rhs` - 1 without allowing overflow
```
kevaundray, line 379: nit: Caculate -> Calculate
nit: cus -> because

stdlib/src/blackbox_fallbacks/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    num_bits + 8 - remainder
+}
+
+pub(crate) fn round_to_nearest_byte(num_bits: u32) -> u32 {
+    round_to_nearest_mul_8(num_bits) / 8
+}
+
+// TODO: Maybe this can be merged with `bit_decomposition`
```
kevaundray, line 27: perhaps move bit_decomposition here too

stdlib/src/blackbox_fallbacks/utils.rs
kevaundray: I think we can instead just pass the VariableStore as a mutable reference. See comment ontop of bit_decomposition as to why we did not do it initially.

kevaundray (commented):
Once we resolve the nits around styling etc, I think we can merge this in under a experimental feature flag, until we get someone to comb through the logic of sha256, so the PR does not get stale. 

We would only need it to be on the method that converts sha256 blackbox functions into regular gates, so tests won't be affected since they will be switched on in acvm/stdlib unconditionally.

acvm/src/compiler/transformers/fallback.rs
```diff
@@ -107,12 +107,25 @@ impl FallbackTransformer {
...
             BlackBoxFuncCall::RANGE { input } => {
                 // Note there are no outputs because range produces no outputs
-                stdlib::fallback::range(
+                stdlib::blackbox_fallbacks::range(
                     Expression::from(input.witness),
                     input.num_bits,
                     current_witness_idx,
                 )
             }
+            BlackBoxFuncCall::SHA256 { inputs, outputs } => {
```
kevaundray, line 116: experimental feature flag would be added here, so that sha256 only gets created when it is turned on

stdlib/Cargo.toml
```diff
@@ -17,3 +17,4 @@ acir.workspace = true
 default = ["bn254"]
 bn254 = ["acir/bn254"]
 bls12_381 = ["acir/bls12_381"]
+testing = ["bn254"]
```
kevaundray, line 20: what is this for?

kevaundray (approved):
I think this is okay to merge since its isolated and not enabled -- great work!

Issue #395, opened 2023-06-21
Remove `Directive` opcodes


TomAFrench:
### Problem

Post https://github.com/noir-lang/noir/issues/1772, the `Directive` opcode will be unused and replaced with Brillig implementations.

### Happy Case

We should remove the `Directive` opcode to avoid carrying technical debt.

### Alternatives Considered

We could maintain this so that other languages building on top of ACIR can get off the ground more easily without full Brillig support.

### Would you like to submit a PR for this Issue?

No

TomAFrench:
Closing as there's only a single directive left so we can just track that.

PR #422, opened 2023-07-11
feat!: add backend-solvable blackboxes to brillig & unify implementations

### LINKED ISSUE #402: Implementation of blackbox functions should be shareable by the ACVM and Brillig_vm
### Problem

The implementation of blackbox functions has to be the same between the brillig_vm and the acvm pwg. Currently there is no shared crate among those dedicated for this task, so we end up having to duplicate the code for hashes and signature verification in the brillig vm. 

### Happy Case

There is a new crate dedicated to this that is consumed by both the acvm and the brillig vm, or other solution that allows to share the code without generating a cyclic dependency.

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

sirasistant:
# Description

## Problem\*

Implements pedersen, schnorr and fixedBaseScalarMul for Brillig and resolves #402

## Summary\*

Context:
 - Right now ACIR depends on the whole Brillig VM
 - If the brillig VM depends on a solver that imports BlackBoxFunc, you get a cyclic dependency
 - Also, the BlackBoxFunctionSolver trait uses OpcodeResolutionError which is ACVM specific

My proposal:
 - Separate the Brillig_vm crate into a brillig crate and a brillig_vm crate (akin to acir/acvm)
   - Acir only depends on brillig crate, does not import the brillig_vm crate
   - ACVM depends on the brillig_vm crate for solving brillig opcodes
 - Add a new crate blackbox_solver
   - exports functions for implementation-fixed blackboxes
   - exports BlackBoxFunctionSolver trait for backend dependant ones (will eventually be removed)
   - exports an enum BlackBoxResolutionError
   - The ACVM OpcodeResolutionError implements a From<BlackBoxResolutionError>
 - Modify Brillig VM to use a &BlackBoxFunctionSolver and use that for pedersen, schnorr and fixed base scalar mul
 - Modify ACVM to pass the solver to the brillig VM
## Additional Context

# PR Checklist\*


sirasistant:
It's a breaking change due to the change that acir now only exports brillig, not the whole brillig_vm, and the change that now BlackBoxFunctionSolver trait returns results with error type BlackBoxResolutionError instead of OpcodeResolutionError

blackbox_solver/src/lib.rs
```diff
@@ -0,0 +1,281 @@
+#![warn(unused_crate_dependencies)]
```
kevaundray, line 1: Can you give this a crate level doc comment, explaining its usage with //!

brillig/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+#![warn(unused_crate_dependencies)]
+#![warn(unreachable_pub)]
+
+//! The Brillig bytecode is distinct from regular [ACIR][acir] in that it does not generate constraints.
+//! This is a generalization over the fixed directives that exists within in the ACVM.
```
kevaundray, line 5: Linking https://github.com/noir-lang/acvm/issues/395, as we will need to update this sentence, once that is completed

blackbox_solver/src/lib.rs
sirasistant: sure!

kevaundray (approved):
Overall looks good to me, one nit about doc commenting the new crates

PR #430, opened 2023-07-12
chore: add cpp codegen for ACIR [DO NOT MERGE]

### LINKED ISSUE #434: Introduce serde-reflection for ACIR cpp codegen
### Problem

We have been utilizing largely manual bindings of ACIR --> C++ for interoperation between Noir and [barretenberg](https://github.com/AztecProtocol/barretenberg/) (Aztec Lab's proving backend written in C++) with the use of e.g. [acvm-backend-barretenberg](https://github.com/noir-lang/acvm-backend-barretenberg), [acir-to-bberg-circuit](https://github.com/vezenovm/acir-to-bberg-circuit).

This has been causing growing pain at interoperation and identification of breaking ACIR changes with the continuous changes introduced at both Noir's and barretenberg's ends.

### Happy Case

Introduce [serde-reflection](https://github.com/zefchain/serde-reflection) into the ACVM tech stack for automated codegen of ACIR in C++.

This should also serve useful down the road when there is the need to codegen ACIR in other languages for integrating Noir with other proving backends written in other languages.

### Alternatives Considered

We can opt for a proper Interface Definition Language (IDL) for even more robust multilingual codegen, but that would require more explorations and engineering efforts.

For now serde-reflection should be sufficient for what is needed. We can revisit IDL down the road when the need arises.

### Additional Context

_No response_

No

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves https://github.com/noir-lang/acvm/issues/434

## Summary\*

This PR adds a quick implementation of `serde_reflection`. Pushing this up for visibility rather to get it merged as is.

`serde_reflection` fails if it has two different enums with the same name so I've had to rename `brillig::Opcode` to `BrilligOpcode`. 

## Additional Context

# PR Checklist\*


acir/acir.cpp
```diff
@@ -0,0 +1,5260 @@
...
+
+    struct Circuit {
+        uint32_t current_witness_index;
+        std::vector<Circuit::Opcode> opcodes;
+        Circuit::PublicInputs public_parameters;
+        Circuit::PublicInputs return_values;
+
+        friend bool operator==(const Circuit&, const Circuit&);
+        std::vector<uint8_t> bincodeSerialize() const;
+        static Circuit bincodeDeserialize(std::vector<uint8_t>);
```
TomAFrench, line 913: bberg would then just have to iterate through the opcodes field and construct a `ConstraintSystem` from this.

kevaundray:
Now that ACVM has been migrated into noir, we should find a way to put the codegen code in here into the acvm repo in noir for when we want to make serialization changes cc @kobyhallx and @TomAFrench

PR #433, opened 2023-07-12
fix: solvable intermediate gates for csat

guipublic:
# Description

## Problem\*

The width reduction creates intermediate variables from an expression. However, if an unknown variable is used for the intermediate one, from the expression, the intermediate is also unknown. In that case the csat generates an arithmetic gate with 2 unknowns and it cannot be resolved in one step.

Resolves 

## Summary\*

We only use 'known' witnesses when generating intermediate variables.
To do this, we track the 'known' witness in a map, after each opcode that the transformation pass processes.

## Additional Context

# PR Checklist\*


acvm/src/compiler/mod.rs
```diff
@@ -26,6 +26,7 @@ pub enum CompileError {
 /// Applies [`ProofSystemCompiler`][crate::ProofSystemCompiler] specific optimizations to a [`Circuit`].
 pub fn compile(
     acir: Circuit,
+    inputs: Vec<Witness>,
```
TomAFrench, line 29: I think we should track this on the `Circuit` struct.

TomAFrench (commented):
Thanks for this! Will do a full review next week.

acvm/src/compiler/mod.rs
guipublic: Done

acvm/src/compiler/mod.rs
```diff
@@ -107,20 +113,101 @@ pub fn compile(
...
+                        }
+                    }
+                    Directive::Log(_) => (),
+                }
+                new_opcode_labels.push(opcode_label[index]);
+                transformed_gates.push(opcode.clone());
+            }
+            Opcode::Block(_) => todo!(),
+            Opcode::ROM(_) => todo!(),
+            Opcode::RAM(_) => todo!(),
```
TomAFrench, line 186: Do these opcodes not write new witness values currently? The solver currently allows them to do so.

acvm/src/compiler/transformers/csat.rs
```diff
@@ -283,24 +331,32 @@ impl CSatTransformer {
...
+
+                // Add intermediate variable as a part of the fan-in for the original gate
+                gate.linear_combinations.push(inter_var);
+                self.mark_solvable(inter_var.1);
+            } else {
+                remaining_mul_terms.push(mul_term);
+            }
         }
 
         // Remove all of the mul terms as we have intermediate variables to represent them now
```
TomAFrench, line 358: This comment is now stale and should be updated.

acvm/src/compiler/transformers/csat.rs
```diff
@@ -318,29 +374,37 @@ impl CSatTransformer {
...
+                    intermediate_variables,
+                    intermediate_gate,
+                    num_witness,
+                );
+                self.mark_solvable(inter_var.1);
+                added.push(inter_var);
+            }
+            //intermediate gate is not full, but the gate still has too many terms
+            if not_full && gate.linear_combinations.len() > self.width {
+                unreachable!("Could not reduce the expresion");
```
TomAFrench, line 401:
```suggestion
                unreachable!("Could not reduce the expression");
```

acvm/src/compiler/transformers/csat.rs
```diff
@@ -318,29 +374,37 @@ impl CSatTransformer {
...
+            if intermediate_gate.linear_combinations.len() > 1 {
+                let inter_var = Self::get_or_create_intermediate_vars(
+                    intermediate_variables,
+                    intermediate_gate,
+                    num_witness,
+                );
+                self.mark_solvable(inter_var.1);
+                added.push(inter_var);
+            }
+            //intermediate gate is not full, but the gate still has too many terms
```
TomAFrench, line 399:
```suggestion
            // The intermediate gate is not full, but the gate still has too many terms
```

acvm/src/compiler/transformers/csat.rs
```diff
@@ -45,6 +80,7 @@ impl CSatTransformer {
         let mut gate =
             self.partial_gate_scan_optimization(gate, intermediate_variables, num_witness);
         gate.sort();
+        self.solvable_expression(&gate);
```
TomAFrench, line 83: I'm not a fan of this usage of `solvable_expression`, it returning a bool implies that we're going to be handling the case where the expression is unsolvable but we don't.

TomAFrench:
> I think we should track this on the `Circuit` struct as a "required witnesses to begin execution" field.

I thought a bit more about this and decided that we should instead track the private inputs to the circuit rather than the full set of witnesses required to start execution. We can then calculate this from private+public parameters.

acvm/src/compiler/transformers/csat.rs
guipublic: Done

acvm/src/compiler/transformers/csat.rs
guipublic: I have renamed it and it does not return a bool anymore.

acvm/src/compiler/mod.rs
guipublic: The stepwise solver is not compatible with them, so it is expected to fail.
I have changed this to unimplemented with an error message.

PR #437, opened 2023-07-14
fix: Remove panic when we divide 0/0 in quotient directive

kevaundray:
# Description

A directive should ideally not panic when we are dividing zero by zero. The input is not trusted and should be constrained, so it is okay to return 0 when we have 0/0.

Related: https://github.com/noir-lang/noir/issues/1802

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/directives/mod.rs
+++ b/acvm/src/pwg/directives/mod.rs
@@ -188,3 +188,31 @@ fn format_field_string(field: FieldElement) -> String {
     }
     "0x".to_owned() + &trimmed_field
 }
+
+#[cfg(test)]
+mod tests {
+    use acir::{
+        circuit::directives::{Directive, QuotientDirective},
+        native_types::{Expression, Witness, WitnessMap},
+        FieldElement,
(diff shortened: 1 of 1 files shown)


acvm/src/pwg/directives/mod.rs
```diff
@@ -58,7 +58,7 @@ fn solve_directives_internal(
                 None => FieldElement::one(),
             };
 
-            let (int_r, int_q) = if pred_value.is_zero() {
+            let (int_r, int_q) = if pred_value.is_zero() || int_b.is_zero() {
```
vezenovm, line 61:
```suggestion
            let (int_r, int_q) = if pred_value.is_zero() || (int_a.is_zero() && int_b.is_zero()) {
```
Don't we want this? This will give us (0, 0) as the remainder and quotient for whenever int_b is zero. We should add a test for when divisor_is_zero but the dividend is not zero.

vezenovm (approved):

PR #442, opened 2023-07-17
chore: Add `serialize-messagepack` feature flag 

### LINKED ISSUE #439: Feature flags for serialisation do not exist
https://github.com/noir-lang/acvm/pull/436 moved messagepack serialisaiton behind a feature flag which doesn't exist and so is impossible to turn on.

We should either add the necessary flags or remove the associated code.

kevaundray:
# Description

## Problem\*

Resolves #439 

## Additional Context

# PR Checklist\*


TomAFrench (changes requested):
The optional dependency needs to be activated as well.

kevaundray:
> The optional dependency needs to be activated as well.

This gets built either way since its a workspace dependency, so I've made it non-optional (I'm not sure it was ever getting ignore due to it being a workspace dep)

TomAFrench (approved):

kevaundray:
Clippy is viewing rmp-serde as being unused due to it being behind a feature flag which is not on by default

Issue #418, opened 2023-07-10
`load_constant` function in `UInt32` currently loads a `u128` instead of a `u32`


Ethan-000:
### Problem

Currently `load_constant` function in `UInt32` currently loads a `u128` instead of a `u32` and treates it as `u32` because we only deal with `u32`s at the moment expect one edge case where we need `2^32` which does not fit in `u32`

This might be related to https://github.com/noir-lang/acvm/issues/417

### Happy Case

This edge case described above can be properly dealt with

### Would you like to submit a PR for this Issue?

Maybe

Issue #417, opened 2023-07-10
`Uinit32` in `stdlib` fallback can be generalized to `u8`, 'u64' and others when needed


Ethan-000:
### Problem

Currently `Uinit32` only supports `FieldElement`s that are treated as `u32` because thats what we need but it can be generalized to `FieldElement`s representing other types

### Happy Case

Nothing breaks and we have a more generic the implementation 

### Would you like to submit a PR for this Issue?

Maybe

PR #445, opened 2023-07-18
feat(stdlib): Add fallback implementation of `Keccak256` black box function

Ethan-000:
# Description

## Problem\*

Implement `keccak` also
Resolves https://github.com/noir-lang/acvm/issues/417
Resolves https://github.com/noir-lang/acvm/issues/418

## Summary\*

417: changed implementation to a macro so that `UInt8` and `UInt64` can be implemented. 
418: added a function for the special case (we want to load a `2^32` constant that does not fit into `32` bits) and changed the `load_constant` function so that it can only load constant of its size (`u32` for `UInt32`)

## Additional Context

# PR Checklist\*


acvm/Cargo.toml
```diff
@@ -44,3 +44,7 @@ unstable-fallbacks = []
 [dev-dependencies]
 rand = "0.8.5"
 proptest = "1.2.0"
+blake2 = "0.10.6"
+sha2 = "0.10.6"
+sha3 = "0.10.6"
```
TomAFrench, line 49: Can you use the implementations which are exposed in the `blackbox_solvers` module?

stdlib/src/blackbox_fallbacks/keccak256.rs
```diff
@@ -0,0 +1,267 @@
...
+    let mut new_gates = Vec::new();
+    let mut new_inputs = Vec::new();
+
+    // Decompose the input field elements into bytes and collect the resulting witnesses.
+    for (witness, num_bits) in inputs {
+        let num_bytes = round_to_nearest_byte(num_bits);
+        let (extra_gates, inputs, updated_witness_counter) =
+            byte_decomposition(witness, num_bytes, num_witness);
+        new_gates.extend(extra_gates);
+        new_inputs.extend(inputs);
```
TomAFrench, line 62:
```suggestion
        let (extra_gates, extra_inputs, updated_witness_counter) =
            byte_decomposition(witness, num_bytes, num_witness);
        new_gates.extend(extra_gates);
        new_inputs.extend(extra_inputs);
```
We can avoid shadowing here.

stdlib/src/blackbox_fallbacks/uint64.rs
```diff
@@ -0,0 +1,183 @@
...
+        witnesses: &[Witness],
+        mut num_witness: u32,
+    ) -> (Vec<UInt64>, Vec<Opcode>, u32) {
+        let mut new_gates = Vec::new();
+        let mut variables = VariableStore::new(&mut num_witness);
+        let mut uint = Vec::new();
+
+        for i in 0..witnesses.len() / 8 {
+            let new_witness = variables.new_variable();
+            let brillig_opcode = Opcode::Brillig(Brillig {
```
TomAFrench, line 16: We should be able to remove this brillig as well.

stdlib/src/blackbox_fallbacks/keccak256.rs
```diff
@@ -0,0 +1,267 @@
...
+
+fn keccak_round(
+    mut a: Vec<UInt64>,
+    round_const: u64,
+    mut num_witness: u32,
+) -> (Vec<UInt64>, Vec<Opcode>, u32) {
+    let mut new_gates = Vec::new();
+
+    // theta
+    let mut array = Vec::new();
```
TomAFrench, line 164:
```suggestion
    let mut array = Vec::with_capacity(5);
```

stdlib/src/blackbox_fallbacks/keccak256.rs
```diff
@@ -0,0 +1,267 @@
...
+        let (new_state_u64, extra_gates, updated_witness_counter) =
+            keccak_round(state_u64, round_constant, num_witness);
+        state_u64 = new_state_u64;
+        new_gates.extend(extra_gates);
+        num_witness = updated_witness_counter;
+    }
+
+    // turn state back to UInt8
+    let state_u64_witnesses: Vec<Witness> = state_u64.into_iter().map(|x| x.inner).collect();
+    let mut state_u8 = Vec::new();
```
TomAFrench, line 142:
```suggestion
    let mut state_u8 = Vec::with(state_u64_witnesses.len());
```

stdlib/src/blackbox_fallbacks/keccak256.rs
```diff
@@ -0,0 +1,267 @@
...
+) -> (Vec<Witness>, u32, Vec<Opcode>) {
+    let mut new_gates = Vec::new();
+    let num_blocks = input.len() / BLOCK_SIZE + 1;
+
+    // pad keccak
+    let (input, extra_gates, mut num_witness) = pad_keccak(input, num_blocks, num_witness);
+    new_gates.extend(extra_gates);
+
+    // prepare state
+    let mut state = Vec::new();
```
TomAFrench, line 90:
```suggestion
    let mut state = Vec::with_capacity(200);
```

stdlib/src/blackbox_fallbacks/keccak256.rs
```diff
@@ -0,0 +1,267 @@
...
+    for _ in 0..200 {
```
TomAFrench, line 91:
```suggestion
    for _ in 0..STATE_NUM_BYTES {
```

Not wedded to the name but I think we should factor out some of the magic constants to be named.

stdlib/src/blackbox_fallbacks/blake2s.rs
```diff
@@ -197,18 +196,22 @@ fn blake2s_compress(
...
+                UInt32::from_witnesses(&mi_bytes, num_witness);
+            new_gates.extend(extra_gates);
+            m.push(mi[0]);
+            num_witness = updated_witness_counter
+        }
     }
 
-    for i in 0..8 {
-        v.push(state.h[i]);
+    unroll! {
```
kevaundray, line 211: This is not needed imo, we don't really care about performance of these fallback functions right now

kevaundray (approved):
Left one comment on pulling in a dependency to unroll

PR #446, opened 2023-07-19
feat: add optimisations to fallback black box functions on booleans

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds special cases to the fallback implementations of logic black box functions. This removes unnecessary an unnecessary `toLeRadix` directive and the resulting dummy witness which just gets constrained to be equal to an existing witness.

## Additional Context

# PR Checklist\*

### DIFF
--- a/stdlib/src/blackbox_fallbacks/utils.rs
+++ b/stdlib/src/blackbox_fallbacks/utils.rs
@@ -23,6 +23,41 @@ pub(crate) fn round_to_nearest_byte(num_bits: u32) -> u32 {
     round_to_nearest_mul_8(num_bits) / 8
 }
 
+pub(crate) fn boolean_expr(expr: &Expression, variables: &mut VariableStore) -> Expression {
+    &mul_with_witness(expr, expr, variables) - expr
+}
+
+/// Returns an expression which represents a*b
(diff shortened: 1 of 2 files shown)


stdlib/src/blackbox_fallbacks/utils.rs
```diff
@@ -23,6 +23,41 @@ pub(crate) fn round_to_nearest_byte(num_bits: u32) -> u32 {
...
 }
 
+pub(crate) fn boolean_expr(expr: &Expression, variables: &mut VariableStore) -> Expression {
+    &mul_with_witness(expr, expr, variables) - expr
+}
+
+/// Returns an expression which represents a*b
+/// If one has multiplicative term and the other is of degree one or more,
+/// the function creates intermediate variables accordingly
+pub(crate) fn mul_with_witness(
```
kevaundray, line 33: We should open an issue for having this be the one implementation of mul_with_witness ,pub visibility, and having the Noir compiler reference this

TomAFrench: Created [here](https://github.com/noir-lang/acvm/issues/450)

kevaundray (approved):

PR #447, opened 2023-07-19
feat!: atomic memory opcodes

guipublic:
# Description

## Problem\*

Add new ACIR opcodes doing atomic memory operation. They allow ACVM to perform stepwise solving for dynamic memory.

Resolves 

## Additional Context

# PR Checklist\*

### DIFF
--- a/acir/src/circuit/opcodes.rs
+++ b/acir/src/circuit/opcodes.rs
@@ -48,30 +48,16 @@ pub enum Opcode {
 }
 
 #[derive(Clone, PartialEq, Eq, Debug)]
-pub enum OpcodeEnum {
-    Arithmetic,
-    BlackBoxFuncCall,
-    Directive,
-    Block,
-    ROM,
-    RAM,
-    Brillig,
+pub enum UnsupportedOpcode {
     MemoryOp,
     MemoryInit,
 }
 
-impl std::fmt::Display for OpcodeEnum {
+impl std::fmt::Display for UnsupportedOpcode {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            OpcodeEnum::Arithmetic => write!(f, "Arithmetic"),
-            OpcodeEnum::BlackBoxFuncCall => write!(f, "BlackBoxFuncCall"),
-            OpcodeEnum::Directive => write!(f, "Directive"),
-            OpcodeEnum::Block => write!(f, "Block"),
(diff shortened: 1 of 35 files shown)


acir/src/circuit/opcodes.rs
```diff
@@ -36,6 +36,44 @@ pub enum Opcode {
...
+        op: MemOp,
+    },
+    MemoryInit {
+        block_id: BlockId,
+        init: Vec<Witness>,
+    },
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum OpcodeEnum {
```
kevaundray, line 51: Why is this needed?

acvm/src/pwg/block.rs
```diff
@@ -30,50 +30,65 @@ impl BlockSolver {
         self.block_value.get(&index).copied()
     }
 
+    pub(crate) fn init(&mut self, init: &[Witness], initial_witness: &WitnessMap) {
```
kevaundray, line 33: can you add a doc comment saying what this is doing and why it is needed?

acvm/src/compiler/transformers/fallback.rs
```diff
@@ -32,6 +32,13 @@ impl FallbackTransformer {
                     acir_supported_opcodes.push(opcode);
                     continue;
                 }
+                Opcode::MemoryInit { .. } | Opcode::MemoryOp { .. } => {
+                    if !is_supported(&opcode) {
+                        return Err(CompileError::UnsupportedOpcode(opcode.to_enum()));
```
kevaundray, line 37: I think this is a bit misleading, since we have Arithmetic as a variant and this should never return the arithmetic variant because Arithmetic is always supported

acir/src/circuit/opcodes.rs
guipublic: This is needed to indicate that an opcode is not supported inside a CompilerError. The Opcode has a too big size to be used directly.

acvm/src/compiler/transformers/fallback.rs
guipublic: I have renamed the enum and make sure it can only be called for these two opcodes.

acir/src/circuit/opcodes.rs
guipublic: I have renamed it, following the later comment.

TomAFrench:
Marking this breaking as it will change the ACIR serialisation

kevaundray (approved):

PR #452, opened 2023-07-25
chore: Improve readability of `BlockSolver.solve_memory_op`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR just does some code reorganisation and renaming of variables to make this function a little clearer.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/block.rs
+++ b/acvm/src/pwg/block.rs
@@ -75,7 +75,7 @@ impl BlockSolver {
             // into this value.
             let value_read_witness = value.to_witness().expect("This should be a witness");
 
-            let value_in_array = self.read_memory_index(memory_index)?;
+            let value_in_array = self.read_memory_index(memory_index);
 
             insert_value(&value_read_witness, value_in_array, initial_witness)
         } else {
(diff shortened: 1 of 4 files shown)


acvm/src/pwg/block.rs
```diff
@@ -6,34 +6,33 @@ use acir::{
...
 impl BlockSolver {
-    fn insert_value(&mut self, index: u32, value: FieldElement) {
+    fn write_memory_index(&mut self, index: MemoryIndex, value: FieldElement) {
         self.block_value.insert(index, value);
     }
 
-    fn get_value(&self, index: u32) -> Option<FieldElement> {
-        self.block_value.get(&index).copied()
+    fn read_memory_index(&self, index: MemoryIndex) -> FieldElement {
+        self.block_value.get(&index).copied().expect("Should not read uninitialized memory")
```
TomAFrench, line 29: I've chosen to retain this as an `expect` for now. We can turn this into a proper error type later.

TomAFrench:
@kevaundray This should be ready to go in now.

kevaundray (approved):
LGTM - One small nit that I added; we likely want to switch `type MemoryIndex = u32;` to be a newtype. This would force us to explicitly type it and make conversions from u64 to u32 more aparent, versus `x as u64 as MemoryIndex`

PR #456, opened 2023-07-26
chore!: Remove block opcode

kevaundray:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/mod.rs
+++ b/acvm/src/pwg/mod.rs
@@ -227,9 +227,6 @@ impl<B: BlackBoxFunctionSolver> ACVM<B> {
                     res => res.map(|_| ()),
                 }
             }
-            Opcode::Block(_) | Opcode::ROM(_) | Opcode::RAM(_) => {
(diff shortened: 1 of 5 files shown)


acvm/src/pwg/mod.rs
```diff
@@ -350,8 +347,8 @@ pub fn default_is_opcode_supported(language: Language) -> fn(&Opcode) -> bool {
     // The ones which are not supported, the acvm compiler will
     // attempt to transform into supported gates. If these are also not available
     // then a compiler error will be emitted.
-    fn plonk_is_supported(opcode: &Opcode) -> bool {
```
kevaundray, line 353: This and the codeblock that it is in, is part of a deprecated feature, so we should remove that in a separate PR and update Noir to no longe rely on these deprecated method

kevaundray:
Closing as Tom is doing this as part of a general interface cleanup, in the same or possibly a different PR

PR #480, opened 2023-08-10
feat: Add `Sync` to `Backend` in `acvm`

skaunov:
# Description

## Problem\*

Needed for threading in <https://github.com/noir-lang/noir/issues/2196>.

Resolves none

## Summary\*

Add `Sync` to `Backend` in `acvm`

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -38,6 +38,7 @@ pub trait Backend:
     + CommonReferenceString
     + Default
     + Debug
+    + Sync
 {
 }
 


Savio-Sou:
Hey there, thanks for submitting the PR!

Could you elaborate a bit on how you plan on tackling https://github.com/noir-lang/noir/issues/2196 and how would this help execute that plan?

skaunov:
Obviously, it would be iterative for me, as I'm fairly new to the code-base. Pls, see https://github.com/noir-lang/noir/compare/master...skaunov:noir:2196-Run-test-cases-for-nargo-test-in-parallel for the first step I have take. [As you can see,](https://github.com/noir-lang/noir/issues/2196#issuecomment-1668010782) the main idea is to feed test cases one-by-one to few `thread` running in parallel.

It was some time ago, I recollect that I tried to just unfold `workspace` to see how it would work, so there's mainly pasted code for index part. And after analyzing available options I couldn't see a way to run a test in a new `thread` without sending a reference to `Backend` to the thread.

If there's any other preferred way, just show how take it! X)

PR #493, opened 2023-08-14
feat!: Switched from OpcodeLabel to OpcodeLocation and ErrorLocation

sirasistant:
# Description

## Problem\*

This PR allows to start locating errors inside brillig functions, by exporting a OpcodeLocation struct that can be used to locate both ACIR and Brillig opcodes. Also updated the ACIR optimizer to return a map of new location to old location to support keeping track of location of brillig opcodes.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acir/src/circuit/mod.rs
+++ b/acir/src/circuit/mod.rs
@@ -7,7 +7,7 @@ use crate::native_types::Witness;
 pub use opcodes::Opcode;
 use thiserror::Error;
 
-use std::{io::prelude::*, str::FromStr};
+use std::{io::prelude::*, num::ParseIntError, str::FromStr};
 
 use flate2::Compression;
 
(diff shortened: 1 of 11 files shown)


acir/src/circuit/mod.rs
```diff
@@ -31,13 +32,58 @@ pub struct Circuit {
...
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum OpcodeLocationFromStrError {
+    #[error("Invalid opcode location string: {0}")]
+    InvalidOpcodeLocationString(String),
+}
+
+impl FromStr for OpcodeLocation {
```
kevaundray, line 60: can you document why this is needed?

acvm/src/compiler/mod.rs
```diff
@@ -26,23 +28,52 @@ pub enum CompileError {
     UnsupportedMemoryOpcode(UnsupportedMemoryOpcode),
 }
 
+fn create_acir_transformation_map(
```
kevaundray, line 31: Can you add a doc comment specifying what this is doing?

acvm/src/compiler/mod.rs
```diff
@@ -26,23 +28,52 @@ pub enum CompileError {
...
+    acir: &Circuit,
+    acir_opcode_positions: Vec<usize>,
+) -> HashMap<OpcodeLocation, OpcodeLocation> {
```
kevaundray, line 34: possibly create a type alias, since `HashMap<OpcodeLocation, OpcodeLocation>` is not the most informative type and it is used in compile

acvm/src/pwg/mod.rs
```diff
@@ -77,20 +77,27 @@ pub enum OpcodeNotSolvable {
     ExpressionHasTooManyUnknowns(Expression),
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default)]
```
kevaundray, line 80: Can you add a doc comment to this?

acvm/src/compiler/mod.rs
sirasistant: Yeah, I opted for a full struct to avoid creating a large map for large circuits, and a method to transform old locations to the new ones

kevaundray (approved):

PR #494, opened 2023-08-14
feat: Expose a `BlackBoxFunctionSolver` containing a barretenberg wasm from `blackbox_solver`

TomAFrench:
# Description

## Problem\*

Resolves the ACVM component of #492 

## Summary\*

This PR embeds a barretenberg binary into `blackbox_solvers` which can then be used by both `acvm_js` and `nargo`.

Due to the fact that spinning up this binary is an async operation, we don't construct it inside the `ACVM` struct automatically but requires it to be injected from outside.

## Additional Context

# PR Checklist\*

BEGIN_COMMIT_OVERRIDE
feat: Expose a `BlackBoxFunctionSolver` containing a barretenberg wasm from `blackbox_solver` (#494)
chore(acvm)!: Remove `BlackBoxFunctionSolver` from `Backend` trait (#494)
END_COMMIT_OVERRIDE


blackbox_solver/Cargo.toml
```diff
@@ -32,6 +32,28 @@ p256 = { version = "0.11.0", features = [
...
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasmer = { version = "3.3", default-features = false, features = [ "js-default" ] }
+getrandom = { version = "0.2", features = [ "js" ]}
+wasm-bindgen-futures = "0.4.36"
+js-sys = "0.3.62"
+
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+getrandom = "0.2"
+wasmer = "3.3"
```
phated, line 51: While debugging this in acvm-backend-bb in the past, I found that it doesn't work quite right. All features for all dependencies are always activated for every target (it seems that it joins them all instead of filtering via the cfg). You probably need to toggle features with other feature flags 🙈

TomAFrench:
This is ready for review now.

blackbox_solver/src/barretenberg/wasm/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+        let u32_bytes = u32::from_be_bytes(i32_bytes);
+
+        self.transfer_to_heap(bytes, u32_bytes as usize);
+        Ok(ptr.into())
+    }
+
+    // pub(super) fn free(&self, pointer: WASMValue) -> Result<(), Error> {
+    //     self.call("bbfree", &pointer)?;
+    //     Ok(())
+    // }
```
kevaundray, line 249:
```suggestion
```

kevaundray (approved):

PR #495, opened 2023-08-16
feat(acvm_js): Add `execute_circuit_with_black_box_solver` to prevent reinitialization of `BlackBoxFunctionSolver`

dbanks12:
# Description

## Problem\*

It was determined that pedersen generator initialization was occurring for every call to `acvm_js`'s `execute_circuit` (if the circuit includes a pedersen op). 

Resolves https://github.com/AztecProtocol/aztec-packages/issues/1553

## Summary\*

1. Renames `SimulatedBackend` to `WasmBlackBoxFunctionSolver`
2. Adds wasmbind to `createBlackBoxSolver`
3. Adds wasmbind to `executeCircuitWithBlackBoxSolver`
    * (accepts object created in 2^)

(keeps the old `executeCircuit` wasmbind to avoid breaking change)

This massively improves runtime of Aztec's `acir-simulator` which makes heavy use of `execute_circuit`.

This PR also includes test node and browser refactors to reflect the API change.

# PR Checklist\*

BEGIN_COMMIT_OVERRIDE
feat(acvm_js): Add `execute_circuit_with_black_box_solver` to prevent reinitialization of `BlackBoxFunctionSolver`
chore(acvm)!: Pass `BlackBoxFunctionSolver` to `ACVM` by reference
BEGIN_COMMIT_OVERRIDE


kevaundray:
I think this would be a breaking change since it modifies the `execute_circuit`

acvm_js/src/execute.rs
```diff
@@ -23,6 +24,11 @@ impl SimulatedBackend {
     }
 }
 
+#[wasm_bindgen(js_name = "createBackend")]
+pub async fn create_backend() -> SimulatedBackend {
```
TomAFrench, line 28: We're not creating a backend here but a `BlackBoxFunctionSolver`. If we're exposing this externally then we should name these correctly.

```suggestion
pub async fn create_black_box_solver() -> WasmBlackBoxFunctionSolver {
```

acvm_js/src/execute.rs
```diff
@@ -71,14 +77,14 @@ impl BlackBoxFunctionSolver for SimulatedBackend {
 /// @returns {WitnessMap} The solved witness calculated by executing the circuit on the provided inputs.
 #[wasm_bindgen(js_name = executeCircuit, skip_jsdoc)]
 pub async fn execute_circuit(
+    backend: &SimulatedBackend,
```
TomAFrench, line 80: Can you revert this change and create a new function which takes an explicit black box function solver? This will make this change non-breaking.

TomAFrench (changes requested):

acvm_js/src/execute.rs
dbanks12: So we should basically rename `SimulatedBackend` itself right?

dbanks12: Renamed `SimulatedBackend` -> `WasmBlackBoxFunctionSolver`

acvm_js/src/execute.rs
dbanks12: Created above-mentioned new function, and then changed `execute_circuit` to just initialize a solver and call `execute_circuit_with_black_box_solver()` to avoid nearly full duplication of the function logic.

dbanks12:
> I think this would be a breaking change since it modifies the `execute_circuit`

Reverted that change as per @TomAFrench's recommendation and added an additional wasmbind function that accepts a solver.

TomAFrench:
Actually, this _is_ breaking as we're modifying the main ACVM struct

PR #498, opened 2023-08-18
feat: embed native black box solver

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR embeds the native barretenberg library to solve black box functions. This requires us to move to nix

## Additional Context

# PR Checklist\*


TomAFrench:
@kevaundray Bundling a static library into ACVM is causing a certain amount of pain in nix to the point where I think it's worth waiting for Zac's pedersen work so we don't have to fix this up and then immediately tear it down again.

PR #500, opened 2023-08-18
chore: Remove unused Directive::Log

Ethan-000:
# Description

## Problem\*

https://github.com/noir-lang/noir/pull/2358#issuecomment-1682597799

## Additional Context

# PR Checklist\*


TomAFrench (changes requested):
This breaks the serialization of `Circuit`. I'd like to perform all of these removals in one go to avoid the need for updating barretenberg multiple times.

PR #503, opened 2023-08-22
feat(acir)!: Add predicate to MemoryOp

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/2133

## Summary\*

We are executing memory operations even under a zero predicate. This leads to situations where the VM will error with index out of bounds even if the memory operation is under a zero predicate. 

## Additional Context

# PR Checklist\*


acvm/src/pwg/memory_op.rs
```diff
@@ -97,9 +110,19 @@ impl MemoryOpSolver {
             // into the memory block.
             let value_write = value;
 
-            let value_to_write = get_value(&value_write, initial_witness)?;
-
-            self.write_memory_index(memory_index, value_to_write)
+            // A zero predicate indicates that we should skip the write operation and zero out
+            // the memory it touches.
```
vezenovm, line 114: I also debated just not performing a write here at all, but was unsure on which was the safer approach

sirasistant:
I think we might need to update this file https://github.com/noir-lang/acvm/blob/master/acir/tests/test_program_serialization.rs to also test serialization of memory ops. I think this PR needs to update the test cases for JS as well, but since the test doesn't use memory ops, it doesn't alert the dev

kevaundray (approved):
We still need to encode the predicate in the circuit which could be done on the bb side. For now, its just a serialization change and no logic will be required in the cpp to be changed (ie it ignores the predicate) and we will open up an issue about predicate encoding -- It could also be done on the Noir side, so it requires a bit more discussion

vezenovm:
> I think we might need to update this file https://github.com/noir-lang/acvm/blob/master/acir/tests/test_program_serialization.rs to also test serialization of memory ops.

I don't see MemoryOps being used anywhere in this test?

TomAFrench:
Alvaro is saying that we should add a test which _does_ use `MemoryOps` to warn if their serialization changes.

vezenovm:
> Alvaro is saying that we should add a test which _does_ use `MemoryOps` to warn if their serialization changes.

Will add a test

vezenovm:
> We still need to encode the predicate in the circuit which could be done on the bb side. For now, its just a serialization change and no logic will be required in the cpp to be changed (ie it ignores the predicate) and we will open up an issue about predicate encoding -- It could also be done on the Noir side, so it requires a bit more discussion

https://github.com/noir-lang/noir/issues/2493

PR #504, opened 2023-08-22
feat: Implement `Char` from `FieldElement`

phated:
# Description

## Problem\*

Resolves 

## Summary\*

While working on https://github.com/noir-lang/noir/pull/2373, I noticed that we have code that converts a `FieldElement` into a `char` so we can collect them into a `String`. This adds a proper `From` implementation so we don't need the custom code like https://github.com/noir-lang/noir/blob/master/crates/noirc_abi/src/lib.rs#L394-L399

## Additional Context

# PR Checklist\*

### DIFF
--- a/acir_field/src/generic_ark.rs
+++ b/acir_field/src/generic_ark.rs
@@ -111,6 +111,15 @@ impl<F: PrimeField> From<i128> for FieldElement<F> {
     }
 }
 
+impl<F: PrimeField> From<FieldElement<F>> for char {
+    fn from(element: FieldElement<F>) -> Self {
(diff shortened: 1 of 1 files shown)


vezenovm (approved):

TomAFrench:
Do we want to turn the assert into an error and have this be a `TryFrom`? (The `unwrap` can be left as is as it should never trigger).

phated:
> Do we want to turn the assert into an error and have this be a TryFrom? (The unwrap can be left as is as it should never trigger).

This package doesn't have errors 🤷

PR #506, opened 2023-08-27
feat: Implement rem operation

### LINKED ISSUE #496: Print Field elements in a prettier format
### Problem

Printing of large field elements is ux unfriendly, making it hard to figure out what the actual number was.

The motivation was this twitter thread: https://twitter.com/recmo/status/1692216107170426919?s=20

Relevant link: https://en.wikipedia.org/wiki/Rational_reconstruction_(mathematics)

### Happy Case

Numbers like -1 mod p and 10/3 mod p are printed as -1 and 10/3

We should ideally implement the rational construction idea mentioned in that tweet thread; essentially finding a rational approximation of a given number using continued fraction expansions.

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

### LINKED ISSUE #4755: Print Field elements in a prettier format
### Problem

Printing of large field elements is ux unfriendly, making it hard to figure out what the actual number was.

The motivation was this twitter thread: https://twitter.com/recmo/status/1692216107170426919?s=20

Relevant link: https://en.wikipedia.org/wiki/Rational_reconstruction_(mathematics)

### Happy Case

Numbers like -1 mod p and 10/3 mod p are printed as -1 and 10/3

We should ideally implement the rational construction idea mentioned in that tweet thread; essentially finding a rational approximation of a given number using continued fraction expansions.

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

Sakapoi:
# Description

## Problem\*

Resolves noir-lang/noir#4755

## Summary\*

Implemented rem operation
Added `division_with_reminder`  function
Added tests to these functions.

## Additional Context

# PR Checklist\*


acir_field/src/generic_ark.rs
```diff
@@ -441,6 +485,40 @@ mod tests {
         let max_num_bits_bn254 = crate::generic_ark::FieldElement::<ark_bn254::Fr>::max_num_bits();
         assert_eq!(max_num_bits_bn254, 254)
     }
+
+    #[test]
```
kevaundray, line 489: what does this print for -1?

Sakapoi: -1 mod 3 is printed as -1
I added more tests to show more results of the function

kevaundray: Thank you, what about 10/3 as mentioned in the issue?

Sakapoi: You mean I have the number 10 and the number 3. I get the number that will be the result of dividing number 10 by number 3 using the '/' operator. Then I need to take the remainder of the division from the result and output it as 10/3?

TomAFrench:
Can you fix the diff? It's not clear what changes this PR is making.

acir_field/src/generic_ark.rs
Sakapoi: I added another function for printing `FieldElement` in a pretty format when the division is performed. Firstly I tried to change the `Div` operation to return the result instead of `FieldElement`. This idea changed logic a lot, so I just added a separate function.

acir_field/src/generic_ark.rs
```diff
@@ -441,6 +561,101 @@ mod tests {
...
         assert_eq!(max_num_bits_bn254, 254)
     }
+
+    mod division_with_reminder {
+        use crate::generic_ark::FieldElement;
+
+        #[test]
+        fn simple_case() {
+            //test for a simple case: 1000 % 3 = 1
+            let mut dividend = FieldElement::<ark_bn254::Fr>::from(1000 as i128);
```
TomAFrench, line 571:
```suggestion
            let mut dividend = FieldElement::from(1000 as i128);
```

Can you use the type alias which has been defined? These changes need to be applied to the rest of the added tests.

TomAFrench (changes requested):

acir_field/src/generic_ark.rs
```diff
@@ -563,94 +563,94 @@ mod tests {
     }
 
     mod division_with_reminder {
-        use crate::generic_ark::FieldElement;
+        type FieldElement = crate::generic_ark::FieldElement<ark_bn254::Fr>;
```
TomAFrench, line 566: This can be imported, we don't need to create it again.

PR #514, opened 2023-08-31
chore(acvm)!: Remove the `Backend` trait

### LINKED ISSUE #196: Expose backend identifier information

### LINKED ISSUE #209: Generalise SmartContract trait
The smart contract trait is a specialisation of what ACVM can provide. Brian correctly noted that it could be possible to transform circuits into anything like WASM, as long as someone implements the transformation.

### LINKED ISSUE #256: Should `preprocess` be two functions: `generate_proving_key` and `generate_verification_key`?
### Problem

Currently, we require a `preprocess` function on the ProofSystem trait that needs to return the proving key and verification key; however the verifier might only need a verification key. Alternatively, backends such as https://github.com/TomAFrench/acvm-backend-groth16 generate the pk and vk at the same time.

### Proposed solution

Figure out if we want to split these or if it would cause too much friction when implementing various backends.

### Alternatives considered

Leave it as is.

### Additional context

See thread at https://github.com/noir-lang/acvm/pull/231#discussion_r1183490054 for more discussion

### Submission Checklist

### LINKED ISSUE #280: Consider separating `Generator` trait from `ProofSystemCompiler`
### Problem

With the introduction of `CommonReferenceString` trait in #231, we now have code related to Generators split across traits (such as `preprocess` on `ProofSystemCompiler`).

### Proposed solution

We should consider reorganizing traits on a Backend for better classification.

### Alternatives considered

Leave them as-is.

### Additional context

This can probably be parked until pluggable backends are implemented. This would probably also include the backends having a way to provide information about the features they support.

### Submission Checklist

TomAFrench:
# Description

## Problem\*

Resolves #280
Resolves #209
Resolves #256
Resolves #196

## Summary\*

This PR removes the `Backend` trait as we will no longer need to communicate through this trait.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/lib.rs
+++ b/acvm/src/lib.rs
@@ -4,10 +4,6 @@
 pub mod compiler;
 pub mod pwg;
 
-use acir::{
-    circuit::{Circuit, Opcode},
-    native_types::WitnessMap,
-};
 pub use blackbox_solver::{BlackBoxFunctionSolver, BlackBoxResolutionError};
 use core::fmt::Debug;
 use pwg::OpcodeResolutionError;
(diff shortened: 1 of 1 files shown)


kevaundray:
196 also has a TODO in nargo, where we have a hardcoded constant for the ccvm-backend-barretenberg

kevaundray (approved):

PR #518, opened 2023-09-02
fix: Add WASI 20 `_initialize` call to `acvm_backend.wasm` binary

kevaundray:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*


acvm_js/test/node/execute_circuit.test.ts
```diff
@@ -205,3 +205,31 @@ it("successfully executes two circuits with same backend", async function () {
   expect(solvedWitness0).to.be.deep.eq(expectedWitnessMap);
   expect(solvedWitness1).to.be.deep.eq(expectedWitnessMap);
 });
+
+it.only("successfully executes 500 circuits with same backend", async function () {
```
kevaundray, line 209: Would be good to keep a test like this for regression purposes without the console.log statements

acvm_js/test/node/execute_circuit.test.ts
```diff
@@ -205,3 +205,29 @@ it("successfully executes two circuits with same backend", async function () {
...
```
LHerskind, line 209: You probably don't want the `only` in here?

kevaundray: Good point, just fixed

LHerskind (approved):
LGTM

PR #520, opened 2023-09-03
chore: Run tests using Rust 1.70 [DO NOT MERGE]

kevaundray:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup toolchain
         uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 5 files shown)


TomAFrench:
Just in case this eventually gets `[DO NOT MERGE]` removed. We should keep the flake+CI testing on our MSRV, it's ok to have test extra versions but we should always make sure to test our MSRV as well.

kevaundray:
This would be contingent on whether wasmer decides to put the patch into version 4 and version 3. version 4 requires us to raise the MSRV

PR #523, opened 2023-09-04
feat!: Provide runtime callstacks for brillig failures and return errors in acvm_js

sirasistant:
# Description

## Problem\*

ACVM part of https://github.com/noir-lang/noir/issues/2550

 - Brillig errors were never emitted from the ACVM, they were changed to an unsatisfied constraint error. Now they are emitted and they contain a call stack instead of a failing opcode, to be able to provide call stacks for brillig functions which are not inlined.
 - Also, the acvm_js was throwing strings instead of errors. Now it throws JS errors and in the case of an execution failure, an error with a call stack property.
 - Since the acvm_js now returns the failing opcode locations in a structured way, we can remove insertion of indices in OpcodeResolutionError

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/brillig.rs
+++ b/acvm/src/pwg/brillig.rs
@@ -1,6 +1,9 @@
 use acir::{
     brillig::{RegisterIndex, Value},
-    circuit::brillig::{Brillig, BrilligInputs, BrilligOutputs},
+    circuit::{
+        brillig::{Brillig, BrilligInputs, BrilligOutputs},
+        OpcodeLocation,
+    },
     native_types::WitnessMap,
     FieldElement,
 };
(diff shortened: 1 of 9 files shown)


acvm/src/pwg/brillig.rs
```diff
@@ -18,6 +21,7 @@ impl BrilligSolver {
         initial_witness: &mut WitnessMap,
         brillig: &Brillig,
         bb_solver: &B,
+        acir_index: usize,
```
kevaundray, line 24: Spoke on huddle, can create an issue or followup PR for this to make it the same with all solvers

kevaundray (approved):

PR #524, opened 2023-09-04
fix: initialize recursive proof output to zero

guipublic:
# Description

## Problem\*

Recursive aggregation opcode is solved by the backend and cannot be solved by acvm. This can lead to missing assignment errors during execution.

Resolves 

To prevent these errors, we solve the opcode with zeros. The backend will resolved it properly after.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/blackbox/mod.rs
+++ b/acvm/src/pwg/blackbox/mod.rs
@@ -1,10 +1,11 @@
 use acir::{
     circuit::opcodes::{BlackBoxFuncCall, FunctionInput},
     native_types::{Witness, WitnessMap},
+    FieldElement,
 };
 use blackbox_solver::{blake2s, keccak256, sha256};
 
-use super::{OpcodeNotSolvable, OpcodeResolutionError};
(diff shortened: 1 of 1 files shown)


acvm/src/pwg/blackbox/mod.rs
```diff
@@ -150,6 +151,13 @@ pub(crate) fn solve(
         BlackBoxFuncCall::FixedBaseScalarMul { input, outputs } => {
             fixed_base_scalar_mul(backend, initial_witness, *input, *outputs)
         }
-        BlackBoxFuncCall::RecursiveAggregation { .. } => Ok(()),
+        BlackBoxFuncCall::RecursiveAggregation { output_aggregation_object, .. } => {
+            // Solve the output of the recursive aggregation to zero to prevent missing assignement errors
```
kevaundray, line 155:
```suggestion
            // Solve the output of the recursive aggregation to zero to prevent missing assignment errors
```

guipublic: Done

kevaundray (approved):

PR #540, opened 2023-09-11
chore!: added benchmarks [DO NOT MERGE]

Sakapoi:
# Description

## Problem\*

Resolves a problem of bench-marking the `acvm` and `acvm_field` crates 

## Summary\*

In order to learn ways to optimize `acvm`, I wrote benchmark tests for all public functions of `acvm` and `acvm_field` crates. The results of these tests can be found in the file `Benchmark_results.md`. A new type was created using the [uint](https://github.com/recmo/uint) library and tests for mathematical operations using this type were also written. Mathematical operations with this new type are much faster than with the ark_bn254 type, so it is necessary to implement the `PrimeField` trait for this new type so that it can be written to the `FieldElement` structure. Should I make a new pool request with this implementation?

## Additional Context

# PR Checklist\*


acir_field/tests/benchmarks.rs
```diff
@@ -0,0 +1,246 @@
...
+    bits_bench, fetch_nearest_bytes_bench, and_bench, xor_bench,
+    add_bench, sub_bench, mul_bench, div_bench
+}
+
+fn add_uint_bench(c: &mut Criterion) {
+    let x: Uint<256, 4> = Uint::from(10000);
+    let y: Uint<256, 4> = Uint::from(77777);
+
+    c.bench_function("add_uint_bench", |b| {
+        b.iter(|| x + y);
```
kevaundray, line 153: This is not doing modular addition, we'd want (x + y) % bn254_modulus

PR #549, opened 2023-09-13
chore(ci): add `cargo deny`

TomAFrench:
# Description

## Problem\*

Resolves 

## Additional Context

# PR Checklist\*

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -1,0 +1,26 @@
+name: deny
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


TomAFrench:
@kevaundray Thoughts on adding this workflow to try and head off issues like wildcard dependencies in future?

kevaundray:
I think this is a good solution to the wildcard situation and other things we may want to deny

kevaundray (approved):

PR #399, opened 2023-06-21
feat(acvm)!: Support stepwise execution of ACIR

TomAFrench:
# Description

## Problem\*

Janky implementation of stepwise execution.

## Summary\*

This PR makes the assumption that each opcode can be executed on the first attempt (excluding foreign calls). We can then implement stepwise execution.

## Additional Context

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #558, opened 2023-09-20
feat!: ACVM no longer takes ownership over opcodes

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

With the changes in #399, the only reason that ACVM needs ownership over the circuit's opcodes is that it needs to be able to write the return values from foreign calls into the `Brillig` opcodes. This doesn't really make sense as this is ACVM state rather than a property of the bytecode.

I've then moved these into a hashmap held on the `ACVM` struct which get injected into the solver when we encounter a brillig opcode.

## Additional Context

# PR Checklist\*

### DIFF
--- a/acvm/src/pwg/mod.rs
+++ b/acvm/src/pwg/mod.rs
@@ -126,7 +126,7 @@ impl From<BlackBoxResolutionError> for OpcodeResolutionError {
     }
 }
 
-pub struct ACVM<'backend, B: BlackBoxFunctionSolver> {
+pub struct ACVM<'opcodes, 'backend, B: BlackBoxFunctionSolver> {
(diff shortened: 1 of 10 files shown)


acvm/src/pwg/mod.rs
```diff
@@ -126,7 +126,7 @@ impl From<BlackBoxResolutionError> for OpcodeResolutionError {
     }
 }
 
-pub struct ACVM<'backend, B: BlackBoxFunctionSolver> {
+pub struct ACVM<'opcodes, 'backend, B: BlackBoxFunctionSolver> {
```
TomAFrench, line 129: Don't think we actually need two lifetimes here.

